PR #19, opened 2021-05-31
Add sample DKMS configuration file

tgurr:
The source directory name has to include the version and match the one
in dkms.conf, e.g. /usr/src/hid-tmff2-0.0 & PACKAGE_VERSION="0.0".

To add: dkms add -m hid-tmff2 -v 0.0
To remove: dkms remove hid-tmff2/0.0 --all
To build: systemctl restart dkms.service

I went for version 0.0 for now as the driver doesn't have any release/version yet.
For https://github.com/cazzoo/hid-tmff2 it can be extended by adding:
```
BUILT_MODULE_NAME[2]="hid-tmt500rs"
DEST_MODULE_LOCATION[2]="/kernel/drivers/hid"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,4 +53,4 @@ wheel.
 
 + Change timer period:
 
-There have been reports that some games work better with a different timer period (see [#11](https://github.com/Kimplul/hid-tmff2/issues/11) and [#10](https://github.com/Kimplul/hid-tmff2/issues/10)). To change the timer period, add `options hid-tmt300rs timer_msecs=NUMBER` in `/etc/modprobe.d/hid-tmt300rs.conf`. The default timer period is 8, but numbers as low as 2 should work alright.
+There have been reports that some games work better with a different timer period (see [#11](https://github.com/Kimplul/hid-tmff2/issues/11) and [#10](https://github.com/Kimplul/hid-tmff2/issues/10)). To change the timer period, create `/etc/modprobe.d/hid-tmt300rs.conf` and add `options hid-tmt300rs timer_msecs=NUMBER` into it. The default timer period is 8, but numbers as low as 2 should work alright.
--- a/dkms.conf
+++ b/dkms.conf
@@ -1,0 +1,17 @@
+PACKAGE_NAME="hid-tmff2"
+PACKAGE_VERSION="0.0"
+AUTOINSTALL="yes"
+
+# By default, DKMS will add KERNELRELEASE to the make command line; however,
+# this will cause the kernel module build to infer that it was invoked via
+# Kbuild directly instead of DKMS. The dkms(8) manual page recommends quoting
+# the 'make' command name to suppress this behavior.
+MAKE[0]="'make'"
+
+# The list of kernel modules.
+#__DKMS_MODULES
+BUILT_MODULE_NAME[0]="hid-tminit"
+BUILT_MODULE_LOCATION="hid-tminit/"
+DEST_MODULE_LOCATION[0]="/kernel/drivers/hid"
+BUILT_MODULE_NAME[1]="hid-tmt300rs"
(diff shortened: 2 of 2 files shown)


Kimplul:
Thanks for the PR. I changed the branch to `dkms` since I'm not familiar with dkms and from what I understand, I would need to add in some kind of install script that moves the sources into the correct directory, so this is not a full feature by itself yet, right?

Not sure why changing the branch also includes cf31a8e in this pull request, maybe I just goofed something. Anyway, thanks.

tgurr:
> Thanks for the PR. I changed the branch to `dkms` since I'm not familiar with dkms and from what I understand, I would need to add in some kind of install script that moves the sources into the correct directory, so this is not a full feature by itself yet, right?

If distributions begin offering the driver via a package they usually take care of installing the stuff into the proper place, a plain kernel driver doesn't need any special handling in this case as it's really just extracting the release source tarball into /usr/src, the release tarball should be complete and include any git submodules as well though.

As for DKMS it takes care of most of the things like building and copying/moving the kernel modules, but it requires the dkms.conf being present in the source directory to know what to do. Then it's as simple as:

1. cd /usr/src
2. git clone --recurse-submodules https://github.com/Kimplul/hid-tmff2 hid-tmff2-0.0
3. dkms add -m hid-tmff2 -v 0.0
4. systemctl restart dkms.service

berarma:
> Thanks for the PR. I changed the branch to `dkms` since I'm not familiar with dkms and from what I understand, I would need to add in some kind of install script that moves the sources into the correct directory, so this is not a full feature by itself yet, right?

If it helps you decide, I just include instructions to install new-lg4ff using DKMS and let the user copy/link the directory into /usr/src. The user can do it any other way they want but DKMS makes it really easy to install/remove and update the kernel without worrying.

> 
> Not sure why changing the branch also includes [cf31a8e](https://github.com/Kimplul/hid-tmff2/commit/cf31a8e37382fc1a100b3d7f35d3825c8e37e188) in this pull request, maybe I just goofed something. Anyway, thanks.

OP should have rebased their changes to the new branch before merging. PRs compare 2 branches and merge the new commits from one to the other. Since that commit was already in the branch OP was working (master) on, it has been taken as a new commit for the PR.

Kimplul:
Thanks for your clarifications. I still went ahead and wrote a quick script to automate the process. I added a script to automate the process a bit, but I suppose users could still do the install manually. I noticed that Debian doesn't have a `dkms.service`, but the object files still seem to be generated. From what I can tell, Debian uses `/etc/kernel/postinst.d/dkms`, though it only triggers when `lkenrel-install` is called. Shoudln't be an issue, but might be nice to know. I'll do a bit more testing on `dkms` and probably merge it into master soon.

> OP should have rebased their changes to the new branch before merging. PRs compare 2 branches and merge the new commits from one to the other. Since that commit was already in the branch OP was working (master) on, it has been taken as a new commit for the PR.

Actually, I just goofed. I had forgot to pull the newest commit to my local repo when I created the new branch, sorry about that.

berarma:
> Thanks for your clarifications. I still went ahead and wrote a quick script to automate the process. I added a script to automate the process a bit, but I suppose users could still do the install manually. I noticed that Debian doesn't have a `dkms.service`, but the object files still seem to be generated. From what I can tell, Debian uses `/etc/kernel/postinst.d/dkms`, though it only triggers when `lkenrel-install` is called. Shoudln't be an issue, but might be nice to know. I'll do a bit more testing on `dkms` and probably merge it into master soon.

I use Debian and the standard procedure for dkms modules works well. I had to put `REMAKE_INITRD="yes"` in my dkms.conf so that the initrd image would be updated when installing the module to replace the in-kernel module, but that shouldn't be an issue here since you aren't replacing an existing module.

Issue #48, opened 2022-10-20
Thrustmaster TX Support


Austin-2:
Hello,

I was looking for support for a Thrustmaster TX wheel. Mine is the Ferrari 458 wheel, but it is technically a Thrustmaster TX.

https://support.thrustmaster.com/en/product/txracingwheel-en/

From my understanding, this wheel is the same as the T300RS, but for Xbox/PC instead of PS4/PC. The drivers on Windows use the same files.

I was able to get the wheel working with tmdriv, but it did not have FFB.

I would love if there would be support for my wheel on Linux!

thanks!

Kimplul:
Hi, I could try adding the wheel ID to my driver, if we're lucky it might work without much tweaking. Although I wouldn't say it's bound to work, as the T500 also uses the same driver installer files but doesn't work the same way as a T300. Not sure about the base, looks very similar but I guess we'll see.

I see you've found the `hid-tminit` repo. Apparently the TX requires a two-step boot process, at least based on https://gitlab.com/her0/tmdrv/-/blob/master/tmdrv_devices/thrustmaster_tx.py. Shouldn't be an issue, but probably requires modifying the driver a bit more than just adding a wheel value as all other wheels are single stage.

Okay, let's try to get the PC mode working. I added the TX to the `hid-tminit` module accept list according to what I found in [tmdrv](https://gitlab.com/her0/tmdrv/-/blob/master/tmdrv_devices/thrustmaster_tx.py). Please install my fork: https://github.com/Kimplul/hid-tminit/tree/master

With the fork installed, please show the `dmesg` output when plugging in the wheel. There should be an error about an unknown ID, which should be one of the values we're missing. Once we have this, we should be able to move on to tackle the second boot stage.

Austin-2:
> Okay, let's try to get the PC mode working. I added the TX to the hid-tminit module accept list according to what I found in [tmdrv](https://gitlab.com/her0/tmdrv/-/blob/master/tmdrv_devices/thrustmaster_tx.py). Please install my fork: https://github.com/Kimplul/hid-tminit/tree/master

First off, thanks for looking into this for me! I am a bit of a linux noob (trying to set this up on my steamdeck).

I cloned your fork into a folder, went into that folder and ran `make` and it returned this error:

> error: implicit declaration of function ‘hid_is_usb’ [-Werror=implicit-function-declaration]
  300 |         if (!hid_is_usb(hdev))
      |              ^~~~~~~~~~
[...]

PR #69, opened 2023-08-23
Add support for Thrustmaster TX

davidedmundson:
This utilises most of the same logic as the T300, but with a separate sequences for the range, open and close.

This is based on a device with an F1 wheel attachment, this might  extending in the future. 

To anyone testing this patch does not include the "tminit" code, only the force feedback. Use tmdrv to initialize the wheel and then this will take over.  tminit patches will come separately.

### DIFF
--- a/hid-tmtx.c
+++ b/hid-tmtx.c
@@ -1,0 +1,323 @@
+// SPDX-License-Identifier: GPL-2.0
+#include <linux/usb.h>
+#include <linux/hid.h>
+#include "hid-tmff2.h"
+
+#define T248_MAX_EFFECTS 16
+#define T248_BUFFER_LENGTH 63
+
+static const u8 setup_0[64] = { 0x42, 0x01 };
+static const u8 setup_1[64] = { 0x0a, 0x04, 0x90, 0x03 };
+static const u8 setup_2[64] = { 0x0a, 0x04, 0x00, 0x0c };
+static const u8 setup_3[64] = { 0x0a, 0x04, 0x12, 0x10 };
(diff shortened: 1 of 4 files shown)


hid-tmtx.c
```diff
@@ -0,0 +1,323 @@
...
+	FF_TRIANGLE,
+	FF_SQUARE,
+	FF_SAW_UP,
+	FF_SAW_DOWN,
+	FF_AUTOCENTER,
+	FF_GAIN,
+	-1
+};
+
+/* TODO: sort through this stuff */
```
davidedmundson, line 53: I tried dumping the HID, it was absolute garbage. It claimed to have 12 axes, which made SDL ignore it as a controller.

I even tried on Windows and it made the debug dumping utility crash.

This was taken from one of the T300s, using the T428 didn't work nicely.

Kimplul: Sounds like the TX is a bit of a troublemaker, ha.

Kimplul:
Thanks, looks good. I'll buy you a beer if you're ever in Finland :)

Merging and pinging https://github.com/Kimplul/hid-tmff2/issues/48, https://github.com/scarburato/hid-tminit/issues/18. Here's hoping this works out of the box for other attachments as well. I'll add some notes about the TX to the README, could you still check which firmware version you're currently using? Just to get a known good version.

davidedmundson:
I'll let you know.

In all seriousness, I contribute to a lot of projects and you've been the most patient and respsonsive maintainer I've dealt with so thanks for your help getting my wheel working. 

I'll try stay subscribed for any issues other TX owners have and sort out the MR for the tminit side.

PR #73, opened 2023-10-03
Refactor Git Tree structure

RaySlash:
## Description
This pull request focuses on to improve code documentation and readability by refactoring the current tree structure. The tree structure has been changed to as follows:
```
🗳️ hid-tmff2
|
|__ 📁 deps
|       |__ 📁 hid-tminit
|
|__ 📁 dkms
|       |__ dkms-install.sh
|       |__ dkms.conf 
|
|__ 📁 docs
|       |__ DRIVER.md
|       |__ FFBEFFECTS.md
|       |__ INTEGRATION.md
|       |__ STRUCTURE.md
|       |__ CONTRIBUTING.md
|       |__ TODO.md
|
|__ 📁 src
|       |__ 📁 tmt300rs 
|       |       |___ hid-tmt300rs.c
|       |       |___ hid-tmt300rs.h
|       |__ 📁 tmt248
|       |       |___ hid-tmt248.c
|       |__ 📁 tmtx
|       |       |___ hid-tmtx.c
|       |__ hid-tmff2.c
|       |__ hid-tmff2.h
|
|__ Kbuild
|__ Makefile
|__ README.md
|__ LICENSE
|__ .gitignore
|__ .gitmodules
```
Please do propose any change you would like to see. The motivation behind the change was to make the repository less clunky and easily readable. The organized folder structure would help people who are looking onto the repository for the first time to find related code/documentation and ourselves in case  of `hid-tmff2` supports wide range of wheels. I also strongly suggest everyone to refrain from using Inappropriate language. No offense, sorry.

## Milestones
The pull request is planning to add the following changes: (Tick as completed)

- [ ] **REFACTORING**
  - [X] Refactor the directory structure
  - [ ] Implement an `.editorconfig` throughout the project
  - [x] Minimize comments on *.c, *.h etc. files
  - [x] Update hid-tminit to https://github.com/scarburato/hid-tminit/commit/9375f6c7d83af5dd6c8b8fe30351d0f36043b20a
  
- [x] **DOCUMENTATION**
  - [X] Add CONTRIBUTING.md
  - [X] Update README.md
  - [X] Update `*.txt` to `*.md`
  - [x] Review `FFBEffects*.md`
  - [x] Review TODO.md
  - [x] Review CONTRIBUTING.md

- [x] **TESTS**
  - [X] Test module build
  - [x] Test dkms module build

PR #74, opened 2023-10-25
dkms: Fix broken script

RaySlash:
Hi,

In the previous PR [#73](https://github.com/Kimplul/hid-tmff2/pull/73) , the directory structure was changed and this caused dkms build to not function as expected. I have edited the `dkms/dkms-install.sh` appropriately and have tested it.

### DIFF
--- a/dkms/dkms-install.sh
+++ b/dkms/dkms-install.sh
@@ -9,7 +9,7 @@ if ! [ $(id -u) = 0 ]; then
 fi
 
 TM_NAME=hid-tmff2
-TM_VER=0.8
+TM_VER=0.81
 DST="/usr/src/${TM_NAME}-${TM_VER}"
 
 mkdir -p "${DST}"


RaySlash:
Version bumped to `0.81` *( I chose `0.81` as the edits did not improve or bring any feature or bug-fixes for the driver itself )* due to it skipping build. Also, the idea of running script other than from script's directory seems odd. The script should be in a parent directory, i.e., the paths should be relative to the scripts source directory. If unhappy with the change, i suggest on moving the `dkms-install.sh` to project root. Thanks. 😄

Kimplul:
Hello, appreciate the input but I the behaviour is intentional, as running the script from the root folder is documented in the README at the moment.

I somewhat disagree with the statement

> Also, the idea of running script other than from script's directory seems odd.

I understand that there might be confusion about paths if one assumes the paths are relative to the script, but pretty much any script you run on the command line that's in your `$PATH` uses paths relative to your current working directory. Sure, this script isn't really intended to be used as a global script and working dir/script dir distinction isn't as important, but it's still the mental model I use when writing scripts and for example my build scripts in https://github.com/Kimplul/kmi/tree/master/scripts use the format, so I'm a bit resistant to change it here.

That being said, if you want to move the script to the root, that's fine by me, I mostly just don't want to have to `cd` into subdirectories :)

RaySlash:
I will leave it as it was for now then. The `README` should be enough. ig. Bumped the version just to pass build and install as dkms complains no change in src.

Kimplul:
Merged, thanks!

PR #91, opened 2024-04-04
Add SteamDeck specific headers dependency

konchunas:
Valve custom kernel shenanigans


Kimplul:
Cheers, good addition.

PR #94, opened 2024-04-21
Add support for TS-XW Racer

yassineimounachen:
Hi, I pretty much copied exactly the commit for TX support and replaced occurrences of TX with TS-XW, 900° with 1080° and that was it. I didn't expect it to work at all but here we are, haha. Thank you so much, you can't imagine how happy I am. You just saved me from going back to my G27.
I used fftest and SuperTuxKart and they seemed to work alright. I am waiting for other proton games to download and I will test some more.
One caveat is (just like for the TX) initializing the wheel with tmdrv is needed beforehand.
Let me know if I need to add adjustments or if there are any ways to mitigate code duplication.
Thanks again @Kimplul and @davidedmundson.

EDIT  I tried the wheel on EA WRC and it works perfectly minus some pedal mapping I had to perform in game settings

### DIFF
--- a/Kbuild
+++ b/Kbuild
@@ -1,2 +1,2 @@
 obj-m := hid-tmff-new.o
-hid-tmff-new-y := src/hid-tmff2.o src/tmt300rs/hid-tmt300rs.o src/tmt248/hid-tmt248.o src/tmtx/hid-tmtx.o
+hid-tmff-new-y := src/hid-tmff2.o src/tmt300rs/hid-tmt300rs.o src/tmt248/hid-tmt248.o src/tmtx/hid-tmtx.o src/tmtsxw/hid-tmtsxw.o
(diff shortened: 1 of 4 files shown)


Kimplul:
Fantastic, nice work!

You may still want to patch Oversteer to add support for the wheel, you can see an example of what needs to be changed here: https://github.com/berarma/oversteer/pull/169/files

Additionally, some games use SDL's internal list of wheels to enable certain features, I think the F1 series does it, so you might also want to add the wheel to https://github.com/libsdl-org/SDL/blob/c623bd2fdcca29c513f5b68e7bd89368929ee564/src/joystick/SDL_joystick.c#L340

Don't really know what happened to the plans for getting the TX to work with `hid-tminit`, @davidedmundson any updates?

davidedmundson:
@yassineimounachen for Proton games you sometimes need to proxy it through protopedal to get support.
`./protopedal --name 'Daves wheel' --vendor 046d --product c262 /dev/input/by-id/usb-Thrustmaster_Thrustmaster_TX_Racing_Wheel-event-joystick`

> Don't really know what happened to the plans for getting the TX to work with `hid-tminit`, @davidedmundson any updates?

Being honest, once I was able to play my games and I forgot all about it. But now I'm reminded I'll try and get back to upstreaming something. Sorry!

PR #95, opened 2024-04-25
Init TX and TS-XW with the tmdrv method but as a kernel module

yassineimounachen:
Hi, I made a barebones driver that can initialize the TS-XW and TX (the latter is untested), effectively making calling tmdrv.py no longer needed for these two wheels. There are no error checks etc. It does work though. The init data were taken from https://gitlab.com/her0/tmdrv (thank you @her001!).

@davidedmundson Can you please test it and/or suggest any changes?

Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,19 @@
 KDIR ?= /lib/modules/$(shell uname -r)/build
 
-all: deps/hid-tminit
+all: deps/hid-tminit deps/usb-tminit
 	$(MAKE) -C $(KDIR) M=$(shell pwd) modules
 
-install: deps/hid-tminit
+install: deps/hid-tminit deps/usb-tminit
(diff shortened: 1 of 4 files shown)


Kimplul:
Appreciate the initiative, but could you add your changes to https://github.com/Kimplul/hid-tminit/tree/usb? It's a fork I made of `hid-tminit` that creates a HID and a USB module for initializing all Thrustmaster wheels, I would prefer having everything related to initialization in a single place. I'm still doing some testing but the plan is to eventually merge it into `hid-tminit`.

Didn't realize the TX/TS-XW wheels were also showing up as USB devices rather than HID devices, I would've mentioned the fork earlier if I had known. Sorry.

yassineimounachen:
OK I will do that soon. Thank you for letting me know.

yassineimounachen:
https://github.com/Kimplul/hid-tminit/pull/1

Issue #46, opened 2022-07-28
Windows drivers inside prefix


Raboebie:
I know this will sounds a bit strange but I have been able to reproduce this on multiple occasions.

I recently acquired F1 22 and it did not recognize my wheel no matter what I did. Eventually I gave up assuming it was a game issue.

A few days later I noticed a post on vales proton bug tracker about someone who installed the windows drivers inside the prefix and then the wheel was detected and mapped correctly on startup. I was pretty sure it won’t do anything, knowing wine has massive limitation when it comes to driver support, but it did. The wheel is detected and works perfectly.

I then decided to test other titles such as ams2 and acc. With both titles the games did not recognize my previous configuration and I had to recalibrate everything. What really surprised me is that the feel of the wheel and especially the pedals changed a lot after installing the drivers. I race daily and I have very fine settings I always keep, but this changed it quite significantly. Can’t really say if it’s for better or worse but the pedals do provide more depth or range of motion for sure.

Could you shed some light on this perhaps? I honestly feel like this should not have done anything yet it made a big difference.

Kimplul:
I wouldn't be surprised if the driver also installs some registry keys or something along those lines at the same time, and the games check for their existence and change their behaviour based on it. This is actually something that was brought up earlier this summer and I'm intending to take a look at it, but haven't managed to do so yet.

Interesting that the driver even installs under Wine though, but pretty cool. I might add a note about it in the README.

Raboebie:
> t looked into it but this is my working theory.

Thanks for the inputs. That is what I thought may be happening but it's cool that it works non the less.

Raboebie:
Did some quick digging and found these reg keys that the driver creates. Interesting that it says T500RS there. I  obviously have a T300RS. I don't know what that binary data means but I think the directory dwords it creates may be what makes it tick.

![1](https://user-images.githubusercontent.com/4992098/181698425-61534f54-8d6a-429e-8499-9114e2b2fe18.png)
![2](https://user-images.githubusercontent.com/4992098/181698430-0243372a-a672-4d2d-ac04-055066c27eb6.png)

Kimplul:
Could be worth trying to write a script that automatically inserts those keys into a prefix and checking if that has an effect on games?

Raboebie:
That's a good idea. I'll give that a go a bit later.

Raboebie:
[...]

PR #105, opened 2024-06-16
Force effect corrections part1

MmAaXx500:
I read #46 and wondered why they can feel it different with the driver installed inside the prefix, I thought maybe the driver applies some modification to the values that got sent out.

To test the things out I installed a Win10 in VirtualBox, installed the TM drivers, and wrote a POC program to send constant force effect with a range of force levels. I have done the same on Linux and Wine and compared the force levels that got sent out using Wireshark.

Some notable differences between the two:

- On Windows, the available range is from `-10000` to `10000` (`DI_FFNOMINALMAX`)
- On Linux, the range is from`-0x8000` to `0x7FFF`

Things I did:
- Wrote a POC for Windows using dinput to send a constant force effect with only the `lMagnitude` (level) set
- Wrote a POC for Linux to send a constant force effect with only the `level` set
- Used the same Windows program under Wine
- I tried to send new effect uploads instead of updates (Wine doesn't like my idea)

Things I found:

- Windows driver maps the `[-10000;10000]` range to `[-16385;16381]`
- Windows driver handles `<-10000` and `>10000` values, and the sent force is `-16385` and `16381` respectively
- Wine doesn't like `<-10000` and `>10000` values, so I left them out
- Wine really likes to send updates (even for unchanged parameters) instead of new effect uploads
- This driver maps the `[-0x8000;0x7FFF]` range to `[-0x8000;0x7FFF]`, so essentially from 50% in either direction the force is unchanged (I don't feel any difference in the force above `16381`)
- The TM drivers are not installing correctly under Wine, so made no difference for me. Maybe the Thrustmaster SDK (`tm_api_lib_*.dll`)  is involved when a difference is made.

Based on these observations I gathered the requested and sent out values under Windows, Wine, and Linux and made a nice plot out of it. I also modified the driver to roughly match the force levels of the Windows driver by dividing the force level by 2.

Plot of the original and the modified force levels:
ffbplot

All the things I said are relevant only to the T300RS, I have no other wheels available.

Checklist:

- [x] Constant force
- [ ] Periodic
- [ ] Square wave
- [ ] Triangle wave
- [ ] Sine wave
- [ ] Saw up wave
- [ ] Saw down wave
- [ ] Ramp
- [ ] Spring
- [ ] Friction
- [ ] Damper
- [ ] Rumble
- [ ] Inertia
- [ ] Gain
- [ ] Autocenter
- [ ] Envelope(s)

If you think it's not a complete nonsense, and it's wroth to go through the checklist, let me know.

POCs
[ffbtest_win.cpp.txt](https://github.com/user-attachments/files/15857657/ffbtest_win.cpp.txt)
[ffbtest_linux.cpp.txt](https://github.com/user-attachments/files/15857660/ffbtest_linux.cpp.txt)

Wireshark captures
[const_-8000-7fff_win.pcapng.gz](https://github.com/user-attachments/files/15857448/const_-8000-7fff_win.pcapng.gz)
[const_-8000-7fff_linux_wheel.pcapng.gz](https://github.com/user-attachments/files/15857449/const_-8000-7fff_linux_wheel.pcapng.gz)
[const_-10000-10000_wine_wheel.pcapng.gz](https://github.com/user-attachments/files/15857450/const_-10000-10000_wine_wheel.pcapng.gz)
[const_-8000-7fff_linux_corr2_wheel.pcapng.gz](https://github.com/user-attachments/files/15857452/const_-8000-7fff_linux_corr2_wheel.pcapng.gz)
[const_-10000-10000_wine_corr2_wheel.pcapng.gz](https://github.com/user-attachments/files/15857453/const_-10000-10000_wine_corr2_wheel.pcapng.gz)

Gathered data
[ffb.ods](https://github.com/user-attachments/files/15857455/ffb.ods)

### DIFF
--- a/src/tmt300rs/hid-tmt300rs.c
+++ b/src/tmt300rs/hid-tmt300rs.c
@@ -344,6 +344,26 @@ static u8 damper_values[] = {
 	0x7f, 0x07
 };
 
+static void t300rs_calculate_periodic_values(struct ff_effect *effect)
+{
+	struct ff_periodic_effect *periodic = &effect->u.periodic;
+
+	effect->replay.length -= 1;
+
+	periodic->magnitude = (periodic->magnitude * fixp_sin16(effect->direction * 360 / 0x10000)) / 0x7fff;
+
+	if (periodic->magnitude < 0){
+		/* the wheel handles positive magnitudes only */
+		periodic->magnitude = -periodic->magnitude;
+
+		/* to give the expected result 180 deg is added to the phase */
+		periodic->phase = (periodic->phase + (0x10000 / 2)) % 0x10000;
+	}
+
+	/* the interval [0; 32677[ is used by the wheel for the [0; 360[ degree phase shift */
+	periodic->phase = periodic->phase * 32677 / 0x10000;
+}
+
 int t300rs_send_buf(struct t300rs_device_entry *t300rs, u8 *send_buffer, size_t len)
 {
 	int i;
@@ -828,25 +848,29 @@ static int t300rs_update_spring(struct t300rs_device_entry *t300rs,
 	return t300rs_update_damper(t300rs, state);
 }
 
-
 static int t300rs_update_periodic(struct t300rs_device_entry *t300rs,
 		struct tmff2_effect_state *state)
 {
 	struct ff_effect effect = state->effect;
 	struct ff_effect old = state->old;
-	struct ff_periodic_effect periodic = effect.u.periodic;
-	struct ff_periodic_effect periodic_old = old.u.periodic;
 	struct __packed t300rs_packet_mod_periodic {
 		struct t300rs_packet_header header;
 		uint8_t attribute;
 		uint16_t value;
 	} *packet_mod_periodic = (struct t300rs_packet_mod_periodic *)t300rs->send_buffer;
 
+	struct ff_periodic_effect periodic, periodic_old;
 	int ret;
-	int16_t magnitude, phase;
+	uint16_t phase;
+	int16_t magnitude;
 
-	magnitude = (periodic.magnitude * fixp_sin16(effect.direction * 360 / 0x10000)) / 0x7fff;
+	t300rs_calculate_periodic_values(&effect);
+	periodic = effect.u.periodic;
+	magnitude = periodic.magnitude;
 	phase = periodic.phase;
+
+	t300rs_calculate_periodic_values(&old);
+	periodic_old = old.u.periodic;
 
 	if ((periodic.magnitude != periodic_old.magnitude)
 			|| (effect.direction != old.direction)) {
@@ -1125,7 +1149,6 @@ static int t300rs_upload_periodic(struct t300rs_device_entry *t300rs,
 		struct tmff2_effect_state *state)
 {
 	struct ff_effect effect = state->effect;
-	struct ff_periodic_effect periodic = state->effect.u.periodic;
 	struct __packed t300rs_packet_periodic {
 		struct t300rs_packet_header header;
 		uint16_t magnitude;
@@ -1138,18 +1161,19 @@ static int t300rs_upload_periodic(struct t300rs_device_entry *t300rs,
 		struct t300rs_packet_timing timing;
 	} *packet_periodic = (struct t300rs_packet_periodic *)t300rs->send_buffer;
 
+	struct ff_periodic_effect periodic;
 	int ret;
-	uint16_t duration, period, offset, periodic_offset;
-	int16_t phase, magnitude;
+	uint16_t duration, period, phase, offset, periodic_offset;
+	int16_t magnitude;
 
-	duration = effect.replay.length - 1;
-	magnitude = (periodic.magnitude * fixp_sin16(effect.direction * 360 / 0x10000)) / 0x7fff;
-
+	t300rs_calculate_periodic_values(&effect);
+	periodic = effect.u.periodic;
+	duration = effect.replay.length;
+	offset = effect.replay.delay;
+	magnitude = periodic.magnitude;
+	period = periodic.period;
 	phase = periodic.phase;
 	periodic_offset = periodic.offset;
-
-	period = periodic.period;
-	offset = effect.replay.delay;
 
 	t300rs_fill_header(&packet_periodic->header, effect.id, 0x6b);
 
--- a/src/tmt300rs/hid-tmt300rs.c
+++ b/src/tmt300rs/hid-tmt300rs.c
@@ -347,6 +347,7 @@ static u8 damper_values[] = {
 static void t300rs_calculate_periodic_values(struct ff_effect *effect)
 {
 	struct ff_periodic_effect *periodic = &effect->u.periodic;
+	int16_t headroom;
 
 	effect->replay.length -= 1;
 
@@ -362,6 +363,11 @@ static void t300rs_calculate_periodic_values(struct ff_effect *effect)
 
 	/* the interval [0; 32677[ is used by the wheel for the [0; 360[ degree phase shift */
 	periodic->phase = periodic->phase * 32677 / 0x10000;
+
+	headroom = 0x7FFF - periodic->magnitude;
+	/* magnitude + offset cannot be outside the valid magnitude range, */
+	/* otherwise the wheel behaves incorrectly */
+	periodic->offset = clamp(periodic->offset, -headroom, headroom);
 }
 
 int t300rs_send_buf(struct t300rs_device_entry *t300rs, u8 *send_buffer, size_t len)
--- a/src/tmt300rs/hid-tmt300rs.c
+++ b/src/tmt300rs/hid-tmt300rs.c
@@ -630,6 +630,8 @@ static int t300rs_update_constant(struct t300rs_device_entry *t300rs,
 	int16_t level;
 
 	level = (constant.level * fixp_sin16(effect.direction * 360 / 0x10000)) / 0x7fff;
+	/* the Windows driver uses the range [-16385;16381] */
+	level = level / 2;
 
 	if ((constant.level != constant_old.level) || (effect.direction != old.direction)) {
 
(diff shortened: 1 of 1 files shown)


Kimplul:
That's really cool, nice work. I suppose since most games are run under Wine, it might make sense to try and match the Wine behaviour to Windows as closely as possible. I would be really interested in seeing how other effects square up, particularly the periodic effect as they seems to be the most popular ones behind the constant force. Sine wave probably the foremost among them.

Unless I'm mistaken, you wrote one program that directly calls DirectInput on Windows/Wine, but did you write the Linux program with direct  ioctl calls or did you use a library like SDL? The approach to handling FFB differs between Windows and Linux and I'm not entirely sure how directly comparable they are, using SDL or something along those lines should (at least in theory) produce the same graphs for the same program regardless of operating system. I wrote a program that does something along those lines, see https://github.com/Kimplul/ffbsdl, but I didn't do anywhere near as neat a job collecting data as you have.

Did you notice any significant difference in games with this fix? I believe at least DiRT Rally only uses the constant effect, and I could image that if the wheel starts clipping forces it could make for a pretty jarring experience.

MmAaXx500:
> I suppose since most games are run under Wine, it might make sense to try and match the Wine behaviour to Windows as closely as possible.

The good thing (at least in case of the constant effect) is matching the Linux driver corrects the behavior inside Wine too, so both native and Wine games should behave correctly.

> Unless I'm mistaken, you wrote one program that directly calls DirectInput on Windows/Wine, but did you write the Linux program with direct ioctl calls or did you use a library like SDL?

The Windows program calls DirectInput the Linux one calls ioctl and write. I attached them as txt to the first post because cpp files cannot be attached for some reason.

> The approach to handling FFB differs between Windows and Linux and I'm not entirely sure how directly comparable they are, using SDL or something along those lines should (at least in theory) produce the same graphs for the same program regardless of operating system.

Yeah, using some library that abstracts away the differences between Linux and Windows would make the graphs match visually, but then I have to deal with whatever the library may modify inside the effects. I tried to be close as possible to the hardware/OS and that was the native API. If I send some effect that way, I can tell what may the driver/OS modified inside them before sent to the wheel. If I have a library between me and the wheel, then I have to deal with that too.

The way I compared them is if something correspond to the maximum force in the native API (`10000` on Windows, `0x7FFF` on Linux) than the command that was sent to the wheel should contain the same force level too.

> Did you notice any significant difference in games with this fix?

I haven't played any games with the wheel recently, so I don't have that "yeah, it feels better" comparison, but numerically it should be correct.
In BeamNG there is an FFB graph UI app that shows what the game wants from the wheel. Around 50% of the requested force, the wheel feels really strong, like 100% force. With this modification, there is no this type of unnecessary strong force feedback.

Kimplul:
> The good thing (at least in case of the constant effect) is matching the Linux driver corrects the behavior inside Wine too, so both native and Wine games should behave correctly.

Good point.

> The Windows program calls DirectInput the Linux one calls ioctl and write. I attached them as txt to the first post because cpp files cannot be attached for some reason.

Oh, completely glossed over those, sorry. Thought `.txt` meant they were data or something. 

> Yeah, using some library that abstracts away the differences between Linux and Windows would make the graphs match visually,

*Might* make them match, or at least that would be the intended behaviour. If there are any differences, we know we're breaking some internal assumption about the Linux FFB subsystem that SDL has, thought that would be a reasonably quick check.

> If I have a library between me and the wheel, then I have to deal with that too.

Granted, that's also a good point.

> I haven't played any games with the wheel recently, so I don't have that "yeah, it feels better" comparison, but numerically it should be correct.
In BeamNG there is an FFB graph UI app that shows what the game wants from the wheel. Around 50% of the requested force, the wheel feels really strong, like 100% force. With this modification, there is no this type of unnecessary strong force feedback.

Makes sense, just wanted to check there weren't any surprises in actual usage. I should probably try out some games I know to work and compare anyway.

Do you want to do more effects in this pull request or do like one effect per PR? I don't think there's any hurry in getting patches out to people but this also seems like something that could be 'rolled out' in steps.

MmAaXx500:
> Do you want to do more effects in this pull request or do like one effect per PR?

If we split it up too much, then there will be a force scaling difference between the effects, so some forces maybe stronger than others. This probably worse than having equally strong forces.
If you know some more frequently used effects, then maybe a PR for the frequent ones and one (or more) for the less frequent ones.

Kimplul:
> If you know some more frequently used effects, then maybe a PR for the frequent ones and one (or more) for the less frequent ones.

Purely anecdotally, the constant effect is the most common one, with the periodic sine a fairly distant second. Autocentering and gain are both common, but in a slightly different way since they're generally act as a global setting. I've seen some games use friction and spring effects, but can't remember ever seeing ramps or other periodic effects than the sinusoidal one. Periodic effects are treated pretty much the same by the wheel so correcting one would likely apply to the others, though.

MmAaXx500:
Part1

- Constant

Part2 (maybe can be 1 too)

- Periodic effects

Part3

- Friction
- Spring
- Autocenter
- Gain

Part4

- Remaining ones

Just an idea, feel free to change anything you see fit.

Kimplul:
Sounds perfectly reasonable either way.

MmAaXx500:
I'm checking the periodic effects now, may take a day or two to complete it.

The constant force is done, the only unknown is the envelopes that the Windows driver places in them if no envelope is added to the effect. Probably irrelevant because they all have 0 length.

I have marked this PR as ready for review if you want to send this out into the wild before I finish the periodic ones.

MmAaXx500:
Pushed the first part of the periodic effect fixes, the remaining part is the offset.

peterlopen:
Amazing work

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -344,6 +344,32 @@ static u8 damper_values[] = {
...
+	if (periodic->magnitude < 0){
+		/* the wheel handles positive magnitudes only */
+		periodic->magnitude = -periodic->magnitude;
+
+		/* to give the expected result 180 deg is added to the phase */
+		periodic->phase = (periodic->phase + (0x10000 / 2)) % 0x10000;
+	}
+
+	/* the interval [0; 32677[ is used by the wheel for the [0; 360[ degree phase shift */
+	periodic->phase = periodic->phase * 32677 / 0x10000;
```
Kimplul, line 365: 32677 is a bit surprising, I would've expected it to be 32768. I guess it could be that the driver is trying to avoid some unwanted behaviour with values at the extreme ranges?

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -948,6 +980,8 @@ static int t300rs_upload_constant(struct t300rs_device_entry *t300rs,
 	int ret;
 
 	level = (constant.level * fixp_sin16(effect.direction * 360 / 0x10000)) / 0x7fff;
+	/* the Windows driver uses the range [-16385;16381] */
```
Kimplul, line 983: The previous comment also applies to this, even more weird is the range being asymmetric.

Kimplul:
Would you mind also writing up a small analysis of your findings and methodology? Doesn't have to be as detailed as the one for the constant force, but at a minimum I'd like to see the test programs you used.

src/tmt300rs/hid-tmt300rs.c
MmAaXx500: I think it maps to some lookup table inside the wheel, although the value choice still odd, but it works out perfectly.
Wine has some issues using the correct values through the whole range, especially from 390 to 399 deg. I don't know why it happens, but it had it before too (see spreadsheet).

src/tmt300rs/hid-tmt300rs.c
MmAaXx500: There is an about `-2` offset on the whole range in the constant force magnitude and in the periodic effect offset.
The Windows driver using `-2` instead of `0` force/offset in these cases and never uses `0` (see spreadsheet).

I have not replicated this because I _think_ it's a rounding error and saw no difference in behavior.

MmAaXx500:
The method was the same: try everything and see what it does.

A comparison of what Windows and what this PR uses:
|                                  | Windows                                         | This PR                             |
|---------------------------| -------------------------------------------- | --------------------------------- |
| Constant level         | `[-16385; 16381]`                            | ` [-16383; 16383] `             |
| Periodic magnitude| `[0; 32764]`                                     | `[0; 32767]`                        |
| Periodic offset         | `[-32767; 32764]`                            | `[-32767; 32767]`               |
| Periodic phase        | `[0; 32677[` in steps of about `91` | `[0; 32677[` in steps of `1` |

Notable things

- The about `-2` offset of Constant level and Periodic offset are not replicated
- This PR uses `0` for `0` input in case of Constant level and Periodic offset. Windows doesn't, it uses `-2`
- Wine does not use correct phase values, especially from 390 to 399 degree phase shift (not because of this PR)

Updated spreadsheet: [ffb.ods](https://github.com/user-attachments/files/15911825/ffb.ods)
Windows test program: [ffbtest_win.cpp.txt](https://github.com/user-attachments/files/15911831/ffbtest_win.cpp.txt)
Linux test program: [ffbtest_linux.cpp.txt](https://github.com/user-attachments/files/15911836/ffbtest_linux.cpp.txt)

src/tmt300rs/hid-tmt300rs.c
Kimplul: 32677 + 91 = 32768, I guess the Windows driver just doesn't use the last interval or something.

> but it works out perfectly.

Does 32768 still work, or did you mean that the 32677 maximum is the highest working value?

MmAaXx500: If we assume the Windows driver is working correctly, then `[0; 32677[ -> [0; 360[` and `[0; 65354[ -> [0; 720[`. 
The wheel responds to the whole `[0, 65535]` interval.

Kimplul:
Played some games (DiRT Rally 1 & 2, ETS2, RRRE) with the corrections, didn't feel any significant difference from before but the data is pretty convincing. I suppose the differences primarily become more noticeable with stronger effects like crashes, which are pretty chaotic by default. Thanks a lot, really interested in seeing what's wrong with the other effects!

PR #110, opened 2024-07-19
Effect corrections part3

MmAaXx500:
Part1: #105
Part2: #107

Planned changes:

 - [x] Ramp
 - [x] Envelope
 - [x] Revise constant and periodic effects

### Ramp

It seems the Windows driver has a bug that is affecting the ramp effect. Changing the invert parameter without changing the duration or the delay causes the effect to stop. I detailed the issue in the docs and found a solution to it.

There is a Wine issue too, the `slope` and `center` or with the old terminology the `difference` and `level` is always `0`, so the `start_level` and the `end_level` must be `0` too. Happening with and without this PR. I'm using wine 9.13 right now (arch package).
Can someone confirm that this happening to them too?
Still not working as of 9.15.

### Envelope

On Windows, the attack/fade_length has a minimum value depending on how far is the target from the attack/fade level. This means that in case of an envelope that goes from zero to max has the minimum length of 256ms.

There is a function that I figured out (in the docs), but choose to not implement it because I see no reason to do it currently. Tested with a constant effect using a 0% to 100% envelope. On Windows the minimum length is 256ms (enforced by the driver) however I tested it on Linux with values of 256, 128, 64, 0 and all of them can be distinguishable just by the feeling of the wheel therefore the wheel handles them.

Spreadsheet: [ffb_p3.zip](https://github.com/user-attachments/files/16731749/ffb_p3.zip) (only the relevant sheets, grown too large)
Windows test program: [ffbtest_win.cpp.txt](https://github.com/user-attachments/files/16731756/ffbtest_win.cpp.txt)
Linux test program: [ffbtest_linux.cpp.txt](https://github.com/user-attachments/files/16731760/ffbtest_linux.cpp.txt)

---

There are still some missing pieces like the trigger buttons and the meaning of some bits. The following are mainly notes for someone who wants to continue the process (maybe future me).

There is an update message for every effect with the header of `49`. The same fields can be found in every init packet (without the header) titled as "effect type" and "update type". During init the update type has the value of `4f` and I suspect there are 2 more fields in there because during update only `41` (duration) and `44` (offset) are observed. Maybe the trigger buttons are hidden in there.

The effect can be started immediately by setting the MSB bit of the `new * effect` field of an init packet. `ea` instead of `6a` in case of constant effect. Using that, at the end of the packet 2 new bytes are appearing in connection with the iteration count of the `IDirectInputEffect::Start` method.

Windows driver can merge multiple updates into one if they were sent close to each other.

### DIFF
--- a/src/tmt300rs/hid-tmt300rs.c
+++ b/src/tmt300rs/hid-tmt300rs.c
@@ -335,6 +335,16 @@ static u8 condition_values[] = {
 	0xff, 0xfe, 0xff
 };
 
+static int16_t t300rs_calculate_constant_level(int16_t level, uint16_t direction)
+{
+	level = (level * fixp_sin16(direction * 360 / 0x10000)) / 0x7fff;
+
+	/* the Windows driver uses the range [-16385;16381] */
+	level = level / 2;
+
+	return level;
+}
+
 static void t300rs_calculate_periodic_values(struct ff_effect *effect)
 {
 	struct ff_periodic_effect *periodic = &effect->u.periodic;
@@ -619,35 +629,6 @@ error:
 	return ret;
 }
 
-static int t300rs_update_simple_duration(struct t300rs_device_entry *t300rs,
-		struct tmff2_effect_state *state, unsigned type)
-{
-	struct ff_effect effect = state->effect;
-	struct __packed t300rs_packet_mod_duration {
-		struct t300rs_packet_header header;
-		uint16_t marker;
-		uint16_t duration;
-	} *packet_mod_duration = (struct t300rs_packet_mod_duration *)t300rs->send_buffer;
-
-	uint16_t duration;
-	int ret = 0;
-
-	duration = effect.replay.length - 1;
-
-	t300rs_fill_header(&packet_mod_duration->header, effect.id, 0x49);
-	packet_mod_duration->marker = cpu_to_le16(0x4100 + type);
-	packet_mod_duration->duration = cpu_to_le16(duration);
-
-	ret = t300rs_send_int(t300rs);
-	if (ret) {
-		hid_err(t300rs->hdev, "failed modifying duration\n");
-		goto error;
-	}
-
-error:
-	return ret;
-}
-
 static int t300rs_update_constant(struct t300rs_device_entry *t300rs,
 		struct tmff2_effect_state *state)
 {
@@ -657,49 +638,54 @@ static int t300rs_update_constant(struct t300rs_device_entry *t300rs,
 	struct ff_constant_effect constant_old = old.u.constant;
 	struct __packed t300rs_packet_mod_constant {
 		struct t300rs_packet_header header;
-		uint16_t level;
+		uint16_t magnitude;
+		struct t300rs_packet_envelope envelope;
+		uint8_t effect_type;
+		uint8_t update_type;
+		uint16_t duration;
+		uint16_t offset;
 	} *packet_mod_constant = (struct t300rs_packet_mod_constant *)t300rs->send_buffer;
-	int ret;
-	int16_t level;
 
-	level = (constant.level * fixp_sin16(effect.direction * 360 / 0x10000)) / 0x7fff;
-	/* the Windows driver uses the range [-16385;16381] */
-	level = level / 2;
+	int ret = 0;
+	int16_t level, old_level;
+	uint16_t length, old_length;
 
-	if ((constant.level != constant_old.level) || (effect.direction != old.direction)) {
+	level = t300rs_calculate_constant_level(constant.level, effect.direction);
+	old_level = t300rs_calculate_constant_level(constant_old.level, old.direction);
 
-		t300rs_fill_header(&packet_mod_constant->header, effect.id, 0x0a);
-		packet_mod_constant->level = cpu_to_le16(level);
+	length = effect.replay.length - 1;
+	old_length = old.replay.length - 1;
+
+	if (level != old_level
+			|| constant.envelope.attack_length != constant_old.envelope.attack_length
+			|| constant.envelope.attack_level != constant_old.envelope.attack_level
+			|| constant.envelope.fade_length != constant_old.envelope.fade_length
+			|| constant.envelope.fade_level != constant_old.envelope.fade_level
+			|| length != old_length
+			|| effect.replay.delay != old.replay.delay) {
+
+		t300rs_fill_header(&packet_mod_constant->header, effect.id, 0x6a);
+		packet_mod_constant->magnitude = cpu_to_le16(level);
+
+		packet_mod_constant->envelope.attack_length =
+			cpu_to_le16(constant.envelope.attack_length);
+		packet_mod_constant->envelope.attack_level =
+			cpu_to_le16(constant.envelope.attack_level);
+		packet_mod_constant->envelope.fade_length =
+			cpu_to_le16(constant.envelope.fade_length);
+		packet_mod_constant->envelope.fade_level =
+			cpu_to_le16(constant.envelope.fade_level);
+
+		packet_mod_constant->effect_type = 0x00;
+		packet_mod_constant->update_type = 0x45;
+		packet_mod_constant->duration = length;
+		packet_mod_constant->offset = effect.replay.delay;
 
 		ret = t300rs_send_int(t300rs);
-		if (ret) {
+		if (ret)
 			hid_err(t300rs->hdev, "failed modifying constant effect\n");
-			goto error;
-		}
-
 	}
 
-	ret = t300rs_update_envelope(t300rs,
-			state,
-			level,
-			effect.replay.length,
-			effect.id,
-			constant.envelope,
-			constant_old.envelope
-			);
-
-	if (ret) {
-		hid_err(t300rs->hdev, "failed modifying constant envelope\n");
-		goto error;
-	}
-
-	ret = t300rs_update_simple_duration(t300rs, state, 0x00);
-	if (ret) {
-		hid_err(t300rs->hdev, "failed modifying constant duration\n");
-		goto error;
-	}
-
-error:
 	return ret;
 }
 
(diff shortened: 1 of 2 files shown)


MmAaXx500:
Hi @Kimplul, I have some questions regarding this snippet, because I have no idea what's happening here.

https://github.com/Kimplul/hid-tmff2/blob/bb4817e4c999f137cae02d9c739580e43b644972/src/tmt300rs/hid-tmt300rs.c#L503-L509

 - Why is multiplication by `duration` and `level` necessary?
 - Why is division by `0x7fff` is required?
 - If these are some scaling magic, I think it's cannot be applied to `*_length` because it makes no sense to scale time.

Kimplul:
IIRC, in FEdit the envelope values were treated as multipliers in the range 0 - 1 of the effect mangitude and duration respectively,  where `0x7fff` is treated as 1. So let's say we have a constant effect with level of 10 and duration of 20, an attack level of `0x3fff` would mean that the force starts out at a level 5 and then rises to 10, not that the force starts out at the maximum value and goes down to 10. Similarly, the attack duration is relative to the whole effect duration, `0x3fff` would in this case be that it takes half the effect to build up from a constant level of 5 to a level 10.

Envelopes seem to be really very rarely used so I wouldn't necessarily be surprised if this isn't actually how it works, it's been ages since I did the envelope stuff. Admittedly I haven't looked at what DInput says about envelopes, nor have I really looked at how other drivers handle them, they've just never really come up anywhere.

MmAaXx500:
Thanks for the explanation, I think it will be easier than this, but the testing is yet to be done. The wheel handles `*_length` as milliseconds and `*_level` as a value between `0` and `0x7fff`, and fortunately they are the same on Linux.  https://elixir.bootlin.com/linux/v6.10/source/include/uapi/linux/input.h#L325

MmAaXx500:
@Kimplul I revised the constant and periodic effects as part of the work on this PR. For both effects, the update can be simplified by sending only one packet, similar to the condition effects in part 2 and the ramp effect in this PR.
Is it okay to include the revised constant and periodic effects in this PR, or do you prefer them to be in a separate one?

Kimplul:
I think it's okay to include in this PR. I presume the changes are fairly small (couple hundred loc or so?) so it's not too difficult to review, and the individual commits will still show up in the history so even if the merge turns out have bugs it'll be fairly easy to pinpoint which part of the merge is broken.

MmAaXx500:
FYI, I will be away for the next few weeks and plan to resume work on the envelopes around mid-August.

MmAaXx500:
Pushed the envelope changes, it's ready now. Updated the first comment to reflect the current state and uploaded the spreadsheet and the test programs.

Kimplul:
Thanks! My schedule is a bit stuffed at the moment, but I'll try to take a closer look by the end of the week. Had a quick glance through and didn't notice anything that stood out too much. My assumption about 100 LoC was apparently a bit off, but no matter. Appreciate the detailed documentation, as always.

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -501,17 +529,23 @@ int t300rs_stop_effect(void *data, struct tmff2_effect_state *state)
 }
 
 static void t300rs_fill_envelope(struct t300rs_packet_envelope *packet_envelope,
-		int16_t level, uint16_t duration, struct ff_envelope *envelope)
+		struct ff_envelope *envelope)
+{
+	// Note: minimal length limitations are not enforced
+	// as per my testing, the wheel can handle lower values well
```
Kimplul, line 535: Nit: I also tend to leave comments of the form "my testing" or "I'm doing something here" but you might want to either differentiate yourself of use passive voice here

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -705,49 +567,44 @@ static int t300rs_update_constant(struct t300rs_device_entry *t300rs,
...
+		t300rs_fill_envelope(&packet_mod_constant->envelope, &constant.envelope);
 
-	if (ret) {
-		hid_err(t300rs->hdev, "failed modifying constant envelope\n");
-		goto error;
-	}
+		packet_mod_constant->effect_type = 0x00;
+		packet_mod_constant->update_type = 0x45;
+		packet_mod_constant->duration = length;
+		packet_mod_constant->offset = effect.replay.delay;
```
Kimplul, line 601: `->duration` and `->offset` has to take byte endianness into account, i.e. use `cpu_to_le16()`. Presumably there aren't too many people gaming on big endian systems, but still.

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -705,49 +567,44 @@ static int t300rs_update_constant(struct t300rs_device_entry *t300rs,
...
 
-		ret = t300rs_send_int(t300rs);
-		if (ret) {
-			hid_err(t300rs->hdev, "failed modifying constant effect\n");
-			goto error;
-		}
+	if (level != old_level
+			|| t300rs_is_envelope_changed(&constant.envelope, &constant_old.envelope)
+			|| length != old_length
+			|| effect.replay.delay != old.replay.delay) {
```
Kimplul, line 591: Nit: I would prefer it if the `if` did an early return if there was nothing to do and the body of the `if` be lifted to function level, that's usually how I tend to format my code blocks.

Also, to make the lines fit a bit better on smaller terminals you might want to move `t300rs_is_envelope_changed()` to be the first thing to check, saves about a tab width.

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -760,64 +617,54 @@ static int t300rs_update_ramp(struct t300rs_device_entry *t300rs,
...
-	difference = ((top - bottom) * fixp_sin16(effect.direction * 360 / 0x10000)) / 0x7fff;
+	length = effect.replay.length - 1;
+	old_length = old.replay.length - 1;
 
+	if (slope != old_slope
+			|| center != old_center
+			|| invert != old_invert
+			|| length != old_length
+			|| effect.replay.delay != old.replay.delay
+			|| t300rs_is_envelope_changed(&ramp.envelope, &ramp_old.envelope)) {
```
Kimplul, line 648: Same nit as in `t300rs_update_constant()`, make `if` do an early return and move `t300rs_is_envelope_changed()` to be checked first.

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -760,64 +617,54 @@ static int t300rs_update_ramp(struct t300rs_device_entry *t300rs,
...
-			|| (effect.direction != old.direction)) {
+		t300rs_fill_envelope(&packet_mod_ramp->envelope, &ramp.envelope);
 
-		t300rs_fill_header(&packet_mod_ramp->header, effect.id, 0x0e);
-		packet_mod_ramp->attribute = 0x03;
-		packet_mod_ramp->difference = cpu_to_le16(difference);
-		packet_mod_ramp->level = cpu_to_le16(level);
+		packet_mod_ramp->effect_type = invert;
+		packet_mod_ramp->update_type = 0x45;
+		packet_mod_ramp->length2 = packet_mod_ramp->length;
```
Kimplul, line 660: Use `cpu_to_le16()` here as well.

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -906,103 +753,58 @@ static int t300rs_update_periodic(struct t300rs_device_entry *t300rs,
...
-			hid_err(t300rs->hdev, "failed modifying periodic phase\n");
-			goto error;
-		}
+	if (periodic.magnitude != periodic_old.magnitude
+			|| periodic.offset != periodic_old.offset
+			|| periodic.phase != periodic_old.phase
+			|| periodic.period != periodic_old.period
+			|| t300rs_is_envelope_changed(&periodic.envelope, &periodic_old.envelope)
+			|| length != old_length
+			|| effect.replay.delay != old.replay.delay) {
```
Kimplul, line 787: Nitto ditto.

src/tmt300rs/hid-tmt300rs.c
```diff
@@ -906,103 +753,58 @@ static int t300rs_update_periodic(struct t300rs_device_entry *t300rs,
...
-		int16_t period = periodic.period;
+		t300rs_fill_envelope(&packet_mod_periodic->envelope, &periodic.envelope);
 
-		t300rs_fill_header(&packet_mod_periodic->header, effect.id, 0x0e);
-		packet_mod_periodic->attribute = 0x08;
-		packet_mod_periodic->value = cpu_to_le16(period);
+		packet_mod_periodic->effect_type = periodic.waveform - 0x57;
+		packet_mod_periodic->update_type = 0x45;
+		packet_mod_periodic->duration = length;
+		packet_mod_periodic->play_offset = effect.replay.delay;
```
Kimplul, line 801: Use `cpu_to_le16()` here as well.

Kimplul (changes requested):
Alright, thank you for your patience. I had a closer look at this and I found a few places where you're not taking endianness into account, I consider those a must fix before I'm willing to merge. I also pointed out a couple style nitpicks that honestly could go one way or the other.

Other than that, everything compiles, games work, I'm very pleased.

src/tmt300rs/hid-tmt300rs.c
MmAaXx500: Is that ok?
"Note: Minimal length limitations are not enforced, as testing shows that the wheel can handle lower values well"

Kimplul: Sounds good.

src/tmt300rs/hid-tmt300rs.c
MmAaXx500: `packet_mod_ramp->length` is already swapped, I don't think `cpu_to_le16()` is needed again.

MmAaXx500:
Addressed all of them except the `packet_mod_ramp->length`. I made a comment about it on the review.

Also fixed an incorrect return type in dcbf4f0.

src/tmt300rs/hid-tmt300rs.c
Kimplul: Oh right, good catch.

Kimplul:
I ran my usual tests and everything seemed to work perfectly fine so I went ahead and merged. I'll probably have a go at testing big endian systems in a little while as well, but I don't think anyone is seriously using them and possible issues (in general, not just with this PR) can be fixed later. I have a MIPS devkit that I haven't used in a while, should be fun getting it out again.

Thanks again, great work!

PR #116, opened 2024-08-17
DKMS module fails to install due to wrong module name

HexSegfaultCat:
Today I've tried to install the DKMS module and had some issues, but I managed to fix it.

The `deps/hid-tminit/Kbuild` generates `hid-tminit-new.ko` file, but current `dkms.conf` makes DKMS look for file `hid-tminit.ko`, so the installation fails.

### DIFF
--- a/dkms/dkms.conf
+++ b/dkms/dkms.conf
@@ -11,7 +11,7 @@ CLEAN[0]="'make' KDIR=\"$kernel_source_dir\" clean"
 
 # The list of kernel modules.
 #__DKMS_MODULES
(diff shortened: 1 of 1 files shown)


Kimplul:
Thanks for the report, knew I'd forgotten something. This patch isn't actually a full fix since we would also want to install `usb-tminit-new` which is the whole reason I changed the module name, so I pushed [1b895a9](https://github.com/Kimplul/hid-tmff2/pull/116/commits/1b895a912ef1e57835609c713fd7dad7c7860964) instead. Still, appreciate it!

PR #128, opened 2024-11-23
fix report_fixup() on linux 6.12

MmAaXx500:
Return type of report_fixup() was changed to `const` in https://github.com/torvalds/linux/commit/fe73965d078670406acee0218f118c0870d6a58b

Fixes: #129


Kimplul:
Thanks!

fuzunspm:
thank you

PR #135, opened 2024-12-28
Update README to mention linuxconsole tools dependencies

foriequal0:
I had a huge dead zone on Euro Truck Simulator 2 although I've installed uvdev rules.
I didn't know I had to install some utilities to make uvdev rules work.

Here's each distro's package repository:
DEB: https://packages.debian.org/bookworm/joystick
AUR: https://archlinux.org/packages/extra/x86_64/joyutils/
RPM: https://src.fedoraproject.org/rpms/linuxconsoletools

I use Ubuntu so I only can confirm `apt` command works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ sudo pacman -S linux-neptune-68-headers
 
 If none of the above work, please do open up an issue.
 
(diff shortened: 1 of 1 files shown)


Kimplul:
Good catch, definitely an oversight on my part. Thanks!

PR #143, opened 2025-01-25
Support for T818

schlegp:
Added support for the T818 and the different settings this wheel has. 
Needs some work still, especially around device selection (I am not sure if the check for version actually differs between T818 and T248) and the RGB endpoint probably also does not follow best practices.

Still wanted to get it out to the world finally so people can contribute.

To make it easier to set the Mode of the wheelbase I have created a fork of oversteer, that also needs some love still:
https://github.com/schlegp/oversteer-t818


schlegp:
@hodasemi Feel free to check it out and see if it works fine on your machine.

hodasemi:
@schlegp 
I tried to build it. I created a PR (https://github.com/schlegp/hid-tmff-dd/pull/2) which primarily updates the submodule reference the same as the master is. The current version with the link to the old reference is not buildable for me.

Kimplul:
Cool! Just for clarification, did you mean for this to be marked WIP or do you want this to get merged now/soon with possible fixes later? "Needs some work still" is a bit ambiguous to me.

src/hid-tmff2.c
```diff
@@ -17,17 +17,17 @@ MODULE_PARM_DESC(timer_msecs,
 		"Timer resolution in msecs");
 
 /* should these be removed and just rely on /sys? */
-int spring_level = 30;
+int spring_level = 0;
```
Kimplul, line 20: I'll do a full review a bit later but just from a cursory glance, do these values need to be changed for this PR? If you think the default values should be different I'd recommend opening up a separate PR and argue your case for why the defaults are wrong.

schlegp:
My rationale was to leave it in WIP until at least one other user (hodasemi) had tested it.

src/hid-tmff2.c
```diff
@@ -43,11 +43,19 @@ MODULE_PARM_DESC(alt_mode,
...
+int gain = 65535;
 module_param(gain, int, 0);
 MODULE_PARM_DESC(gain,
 		"Level of gain (0-65535)");
 
+int mode = 3;
+module_param(mode, int, 0);
+MODULE_PARM_DESC(mode, "FFB Mode of the Wheel (0-3; 0: Comfort, 1: Sport, 2: Performance, 3: Extreme)");
+
+int color = 0xFF26C2FF;
```
Kimplul, line 55: Nit: hex should be lowercase.

src/hid-tmff2.h
```diff
@@ -22,7 +24,7 @@ extern int alt_mode;
  * space out the interrupts so that they all leave at regular intervals, but
  * for now this is good enough, go slow enough that everything works.
  */
-#define DEFAULT_TIMER_PERIOD	8
+#define DEFAULT_TIMER_PERIOD	3
```
Kimplul, line 27: Same, if you think the default should be changed, please open up a separate PR unless this is required for the T818 to work properly.

src/hid-tmff2.c
schlegp: Things like these are exactly what I meant with "Needs some work still" sorry for being so vague there.
For now this is hacked together to get a good experience with the T818 out of the Box, but not every value is properly integrated.
Just wanted to make it available to people before it rots away on my disk.

src/hid-tmff2.h
```diff
@@ -86,10 +90,13 @@ struct tmff2_device_entry {
 	int (*close)(void *data, int);
 	int (*set_gain)(void *data, uint16_t gain);
 	int (*set_range)(void *data, uint16_t range);
+	int (*set_mode)(void *data, uint);
+	int (*set_color)(void *data, uint32_t rgba);
 	/* switch_mode is required to not do anything if we're alredy in the
 	 * specified mode */
 	int (*switch_mode)(void *data, uint16_t mode);
-	ssize_t (*alt_mode_show)(void *data, char *buf);
```
Kimplul, line 92: Wonder why this line was marked removed. Spaces versus tabs, maybe?

src/tmt248/hid-tmt248.c
```diff
@@ -333,4 +333,4 @@ int t248_populate_api(struct tmff2_device_entry *tmff2)
 	tmff2->wheel_destroy = t248_wheel_destroy;
 
 	return 0;
-}
+}
```
Kimplul, line 336: Please give me back my newline :(

udev/99-thrustmaster.rules
```diff
@@ -18,9 +18,9 @@ KERNEL=="hidraw*", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b66d", MODE="0660
 SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b66d", RUN+="/usr/bin/evdev-joystick --evdev %E{DEVNAME} --deadzone 0"
 SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b66d", RUN+="/usr/bin/jscal -s 6,1,1,32767,32767,16384,16384,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,0,0,0,536870912,536870912,1,0,0,0,536870912,536870912 /dev/input/js%n"
 
-# T248 + T128
+# T248 + T128 + T818
 KERNEL=="hidraw*", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b696", MODE="0660", TAG+="uaccess"
-SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b696", RUN+="/usr/bin/evdev-joystick --evdev %E{DEVNAME} --deadzone 0"
+SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b696", RUN+="/usr/bin/evdev-joystick --evdev %E{DEVNAME} --deadzone 0 --fuzz 16"
```
Kimplul, line 23: What does fuzz do?

Kimplul:
> My rationale was to leave it in WIP until at least one other user (hodasemi) had tested it.

Okay, so since we're past that stage you'd want for this to be merged, possibly with some changes per review requests? Sounds good, I'll have a closer look at some point but feel free to start with the nitpicks I pointed out.

udev/99-thrustmaster.rules
schlegp: Fuzz sets the minimum change that gets reported to the system by evdev. The default value is very high, which leads to issues during precise inputs. Might be useful for other wheels aswell.
This is the only documentation I could find explaining it:
https://python-evdev.readthedocs.io/en/latest/apidoc.html#evdev.device.AbsInfo.fuzz

schlegp:
> > My rationale was to leave it in WIP until at least one other user (hodasemi) had tested it.
> 
> Okay, so since we're past that stage you'd want for this to be merged, possibly with some changes per review requests? Sounds good, I'll have a closer look at some point but feel free to start with the nitpicks I pointed out.

Exactly that's the plan, so I will start on the Nitpicks directly.
I did some research around the RGB endpoint and think this might belong in "/sys/class/leds", so I will check how other RGB tools expect it.

The biggest unknown for me is if the device selection makes sense, as the USB Vendor and Product are the same for all new Thrustmaster devices (T128, T248, T818), but the DD wheel differs completely in terms of range and settings.

Kimplul:
> The biggest unknown for me is if the device selection makes sense

Probably not, but that's on me. I've kinda suspected the current method is a bit limited for a while, but so far there hasn't really been a significant need for anything better.

Maybe have a look at `t300rs_get_attachment()`, there's a command that queries some data from the wheel, it *might* be useful for other wheels to figure out what exact model that they are. I haven't really looked into what all the fields are so I've left most of them named something generic, but there is one field that I suspect is the *actual* model of the wheel, but I can't remember what format the value uses so that's something you should check for the T818, I can have a look at my wheels on my end.

udev/99-thrustmaster.rules
Kimplul: Interesting, can't say I've noticed any significant precision issues when playing but admittedly I'm more of a casual sim driver, so I maybe wouldn't be 'in tune' enough with everything to notice. I would recommend applying this to the other wheels as well, preferably in a new PR so it's easier to do a direct comparison.

Nice catch, though!

schlegp: I noticed it when going on long straights in an LMDh in Automobilista 2. 
When turning the wheel only just a little the ingame wheel would not move. This made it really hard to keep the car going in a straight line.

I will create a separate PR so you can check if this is also relevant for the belt driven wheels.

src/tmt248/hid-tmt248.c
schlegp: Newline returned.

src/hid-tmff2.h
schlegp: Changed it back to its original value. I think this change was made before I found out about fuzz.

src/hid-tmff2.c
schlegp: Fixed.

src/hid-tmff2.c
schlegp: Changed it back to its original value.

schlegp:
> Maybe have a look at `t300rs_get_attachment()`, there's a command that queries some data from the wheel, it _might_ be useful for other wheels to figure out what exact model that they are. I haven't really looked into what all the fields are so I've left most of them named something generic, but there is one field that I suspect is the _actual_ model of the wheel, but I can't remember what format the value uses so that's something you should check for the T818, I can have a look at my wheels on my end.

For my wheel this function returns `10` for the model. The Ferrari 296 GT3 wheel seems to be attachment `10`, while the T300 round wheel reports as `6` (although that might just be the ID of the adaptor).
Do you get different model IDs for your wheels?

Kimplul:
> For my wheel this function returns 10 for the model.

My T248 returns
```
model      -> 10
attachment -> 0
field0     -> 2051
field1     -> 1
```

Darn, too bad about `model` not being what I expected. I guess it might still be useful to look at `model + attachment`, though I'm still not sure what the other fields are.

hodasemi:
I can confirm that it compiles now and it works as expected.

src/hid-tmff2.h
Lawstorant: Trailing whitespace

udev/99-thrustmaster.rules
Lawstorant: At hid-universal-pidff we found that for steering, the best value is honestly just 0. For other axes, we're using 8.

Ghoelian: @schlegp how did you get the wheel to work with Automobilista 2? I can't get any game that runs through proton to recognise the inputs from the wheel, but they do see the controller is there oddly.

Lawstorant: Actually, why isn't fuzz and flat co trolled by the driver instead of udev rules that run external programs? It can be changed when hooking up to `.input_configured`

Kimplul: Frankly, because I just didn't know you could do that :) Although now I  wonder what the relationship between `input_set_abs_params` and `joydev`/`evdev` deadzone/fuzz etc. is, from some issues that have been opened it seems that they're not all the same and some application needed the deadzone tweaked via `evdev-joystick` while others needed it be done via `jscal`. Does `joydev` or `evdev` have some extra handling on top of the kernel layer?

MmAaXx500: > I noticed it when going on long straights in an LMDh in Automobilista 2. When turning the wheel only just a little the ingame wheel would not move. This made it really hard to keep the car going in a straight line.
> 
> I will create a separate PR so you can check if this is also relevant for the belt driven wheels.

@schlegp This is present on T300RS too. I noticed in RRRE in straights. Do you mind if I open a PR to turn off fuzz/flat for this driver?

MmAaXx500: @Kimplul The default fuzz/flat is set in [`hid-input.c:1404`](https://elixir.bootlin.com/linux/v6.15.8/source/drivers/hid/hid-input.c#L1404). Fuzzing is done in [`input.c:70`](https://elixir.bootlin.com/linux/v6.15.8/source/drivers/input/input.c#L70). `joydev` reads fuzz/flat in [`joydev.c:973`](https://elixir.bootlin.com/linux/v6.15.8/source/drivers/input/joydev.c#L973) and uses it in [`joydev.c:76`](https://elixir.bootlin.com/linux/v6.15.8/source/drivers/input/joydev.c#L76). `evdev` probably does its filtering in `libevdev`, but not sure. I'm not aware of any extra filtering over these (although I was not aware of any of these even a few days ago). Turing of fuzz/flat using `input_set_abs_params` affects both `jstest` and `evtest`.

PR #156, opened 2025-04-20
Adds reminder to add MOK certificate if using Secure Boot + DKMS.

danielpokladek:
I have just been caught off by this, it might be worth having this as a note/reminder for other users.


Kimplul:
Thanks! The README has gathered quite a few notes, might make sense to write a proper wiki page for them at some point, but this will do nicely until then.

Issue #39, opened 2022-03-30
Race Room Racing Experience


Kimplul:

ddb1024:
I tried RRRE also with the T300.
And see the behavior mentioned by tperka (https://github.com/Kimplul/hid-tmff2/issues/21#issuecomment-1082474483) 
> after car starts it get pulled to the right and wants to stay there

Tried different Proton Versions, Settings... With ge-6.19-2 there isn't this problem but also no FFB at all.

Kimplul:
Okay, I got the game to run and seems like the issue seems to be that the game uploads a periodic effect with magnitude 0, i.e. essentially just a fancy constant effect. The direction of the force is handled with the `offset` parameter, although curiously I can only see negative values. I'm not sure but this seems like a Proton bug to me, as I can replicate this behaviour on Windows by applying effects with the same parameters as Proton tries to feed my driver. This coupled with some autocentering probably explains the 'sticking to the right', although I'm not entirely sure what is meant by 'inverted FFB' as all FFB just sort of stops after a couple seconds for me.

tperka:
Thanks for providing useful info @Kimplul. Will try to create an issue on proton-ge mentioning issues you found. You can find some details about 'inverted FFB' here: https://forum.sector3studios.com/index.php?threads/cant-drive-in-a-straight-line.7674/

Generally, it seems like it could work if instead of pulling to one side it would pull to the center as expected

Raboebie:
Perhaps #46  could help you out here? Would be curious

fuzunspm:
> Okay, I got the game to run and seems like the issue seems to be that the game uploads a periodic effect with magnitude 0, i.e. essentially just a fancy constant effect. The direction of the force is handled with the `offset` parameter, although curiously I can only see negative values. I'm not sure but this seems like a Proton bug to me, as I can replicate this behaviour on Windows by applying effects with the same parameters as Proton tries to feed my driver. This coupled with some autocentering probably explains the 'sticking to the right', although I'm not entirely sure what is meant by 'inverted FFB' as all FFB just sort of stops after a couple seconds for me.

I have the same sticking to the right issue

Kimplul:
Partial update: https://github.com/Kimplul/hid-tmff2/issues/47 led me to find out I had a bug in how I handled durations of effects. I've patched the issue and

> all FFB just sort of stops after a couple seconds for me.

[...]

PR #166, opened 2025-07-24
fix periodic effect length on update

MmAaXx500:
Avoid subtracting twice from the length of the effect. First time in `t300rs_calculate_periodic_values` and
second time in `t300rs_update_periodic`.

This fixes RRRE's sine effect, as discussed in #39 .


Kimplul:
Thanks!

PR #167, opened 2025-07-24
refactor effect length calculation

MmAaXx500:
This is optional and depends on #166.

My aim with this is to remove the scattered `- 1`s and move the effect length calculation to one place, as discussed in #39.
As a side effect, all effects will now last 1ms longer than before, making it the same length as the requested effect length.


Kimplul:
Thanks for this as well. I thought you mean that the duration calculation should be done once for all effects, but this works as well of course.

PR #175, opened 2025-11-03
feat(t500rs): add USB interrupt-transfer driver; Proton/Steam working…

cazzoo:
… (#6)

- Add T500RS-specific USB interrupt transfer driver: src/tmt500rs/hid-tmt500rs-usb.c

- Wire into build system (Kbuild/Makefile)

- Extend core hid-tmff2.{c,h} for device profile + FFB plumbing, including set_gain so in-game gain adjusts force strength correctly

- Update udev rules for proper permissions/identification

- Add tools/build-reload.sh to speed local build+reload during development

Validated with Proton (Steam) titles: input + basic FFB operate as expected.


udev/99-thrustmaster.rules
```diff
@@ -27,3 +27,29 @@ SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b692", RUN+="/us
 # TSPC
 SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b689", RUN+="/usr/bin/evdev-joystick --evdev %E{DEVNAME} --deadzone 0"
 SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b689", RUN+="/usr/bin/jscal -s 6,1,1,32767,32767,16384,16384,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,0,0,0,536870912,536870912,1,0,0,0,536870912,536870912 /dev/input/js%n"
+
+# T500RS initialization mode
```
Kimplul: The init mode is shared between pretty much all Thrustmaster wheels, and is rather useless as an input device. Remove these rules, they're not T500-specific and not really useful even if they're applied to all wheels.

udev/99-thrustmaster.rules
```diff
@@ -27,3 +27,29 @@ SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b692", RUN+="/us
...
+# Find your event number with: cat /proc/bus/input/devices | grep -A 5 "Thrustmaster"
+KERNEL=="hidraw*", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", MODE="0666", TAG+="uaccess"
+KERNEL=="event*", SUBSYSTEMS=="usb", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", MODE="0666", TAG+="uaccess"
+KERNEL=="js*", SUBSYSTEMS=="usb", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", MODE="0666", TAG+="uaccess"
+SUBSYSTEM=="hid", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", MODE="0666", TAG+="uaccess"
+SUBSYSTEM=="hid", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", MODALIAS=="hid:b0003g*v0000044Fp0000B65E", RUN+="/bin/sh -c 'echo %k > /sys/bus/hid/drivers/tmff2/bind'"
+SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", RUN+="/usr/bin/evdev-joystick --evdev %E{DEVNAME} --deadzone 0"
+SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", RUN+="/usr/bin/jscal -s 6,1,1,32767,32767,16384,16384,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,0,0,0,536870912,536870912,1,0,0,0,536870912,536870912 /dev/input/js%n"
+
+# T500RS sysfs attributes permissions (for GUI and non-root access)
```
Kimplul: Oversteer would probably be a better place to put these sysfs rules, that's where all other wheels have them, including other manufacturers.

udev/99-thrustmaster.rules
```diff
@@ -27,3 +27,29 @@ SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b692", RUN+="/us
...
+SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65d", RUN+="/usr/bin/evdev-joystick --evdev %E{DEVNAME} --deadzone 0"
+SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65d", RUN+="/usr/bin/jscal -s 6,1,1,32767,32767,16384,16384,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,0,0,0,536870912,536870912,1,0,0,0,536870912,536870912 /dev/input/js%n"
+
+# T500RS PC mode
+# IMPORTANT FOR PROTON/STEAM GAMES:
+# If you have multiple joystick devices (js0, js1, etc.) that cause conflicts,
+# use the event device directly in Steam launch options:
+#   SDL_JOYSTICK_DEVICE=/dev/input/eventXX %command%
+# Find your event number with: cat /proc/bus/input/devices | grep -A 5 "Thrustmaster"
+KERNEL=="hidraw*", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", MODE="0666", TAG+="uaccess"
```
Kimplul: Other wheels only use the two last rules, the ones that run `evdev-joystick` and `jscal` (and even they are probably kind of unnecessary, see discussion around https://github.com/Kimplul/hid-tmff2/pull/143#discussion_r1986430665, deadzone might fit into this as well)

Do these other rules do anything relevant? `uaccess` is set over in `71-steamdeck.rules` due to apparently the SteamDeck requiring it, but only on `hidraw`, which these rules don't seem to set. If not, I'd just remove them to keep things simple.

Have you checked that the `jscal` rule works? The magic string of numbers seems to be identical to most other ones so I presume it's fine, but magic can sometimes backfire. Something like `jstest-gtk` could show you if you have a deadzone in applications that read from the `/dev/input/js*` files.

src/hid-tmff2.h
```diff
@@ -108,6 +119,9 @@ int tsxw_populate_api(struct tmff2_device_entry *tmff2);
 #define TMT300RS_PS3_ADV_ID	0xb66f
 #define TMT300RS_PS4_NORM_ID	0xb66d
 
+#define TMT500RS_INIT_ID	0xb65d
```
Kimplul: The init ID of `b65d` is shared between all Thrustmaster wheels, not just the T500, and is not useful for this driver.

src/hid-tmff2.h
```diff
@@ -92,6 +98,10 @@ struct tmff2_device_entry {
 	ssize_t (*alt_mode_show)(void *data, char *buf);
 	ssize_t (*alt_mode_store)(void *data, const char *buf, size_t count);
 	int (*set_autocenter)(void *data, uint16_t autocenter);
+		int (*set_spring_level)(void *data, u8 level);
```
Kimplul: `set_spring_level`, `set_damper_level` and `set_friction_level` are not used anywhere as far as I can grep things. All effects should use the global `spring_level`, `damper_level` and `friction_level` variables anyway, which are set by `hid-tmff2.c`

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  }
+
+  T500RS_DBG("Range set to %u degrees (final value=0x%04x)\n", range,
+             target_value);
+
+  return 0;
+}
+
+/* Set spring level (0-100) - called from sysfs */
+static int t500rs_set_spring_level(void *data, u8 level) {
```
Kimplul: Remove, spring level is handled by `hid-tmff2.c`.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+
+  T500RS_DBG("Spring level set to %u%%\n", level);
+
+  /* Note: The new level will be applied when the next spring effect is
+   * uploaded/updated */
+  return 0;
+}
+
+/* Set damper level (0-100) - called from sysfs */
+static int t500rs_set_damper_level(void *data, u8 level) {
```
Kimplul: Remove, damper level is handled by `hid-tmff2.c`.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+    goto err_buffer;
+  }
+
+  /* Initialize per-effect levels to 100% (0-100 range) */
+  t500rs->spring_gain = 100;
+  t500rs->damper_gain = 100;
+  t500rs->friction_gain = 100;
+  t500rs->inertia_gain = 100;
+
+  /* Initialize current range to default (900°) for smooth transitions */
```
Kimplul: 'smooth transitions'? What transitions?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  init_buf[0] = 0x41;
+  init_buf[1] = 0x00;
+  init_buf[2] = 0x00;
+  init_buf[3] = 0x00;
+  ret = t500rs_send_usb(t500rs, init_buf, 4);
+  if (ret) {
+    hid_warn(t500rs->hdev, "Init command 9 (0x41 clear) failed: %d\n", ret);
+  }
+
+  /* Report 0x40 - Final setup (4 bytes) */
```
Kimplul: Not so final, it seems like :p

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+    return len == 2; /* global gain */
+  case 0x0a:
+    return (len == 15) && s == 0x04; /* config */
+  default:
+    return 0;
+  }
+}
+
+/* Debug logging helper (requires local variable named 't500rs') */
+#define T500RS_DBG(fmt, ...)                                                   \
```
Kimplul: Nitpick: I dislike hidden variable name assumptions, prefer taking the `t500rs` variable name explicitly, like `T500RS_DBG(t500rs, ...`.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  ret = t500rs_send_usb(t500rs, buf, 9);
+  if (ret) {
+    hid_err(t500rs->hdev, "Failed to send Report 0x02: %d\n", ret);
+    return ret;
+  }
+
+  /* Report 0x01 - Main effect upload - MATCH WINDOWS DRIVER EXACTLY! */
+  memset(buf, 0, 15);
+  buf[0] = 0x01;
+  buf[1] = 0x00; /* Effect ID 0 (Windows uses 0, not effect->id) */
```
Kimplul: I don't fully know how the T500 operates, but based on this comment it seems like uploading a constant force always overwrites what's in effect slot 0. What if a user uploads multiple constant effects?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+
+  /* Scale to device range (0-127) */
+  right_strength = (right_strength * 127) / 65535;
+  left_strength = (left_strength * 127) / 65535;
+
+  T500RS_DBG("Upload condition: id=%d, type=0x%02x, gain=%u%%, R=%d, L=%d\n",
+             effect->id, effect_type, effect_gain, right_strength,
+             left_strength);
+
+  /* Report 0x05 - Condition parameters (coefficients) */
```
Kimplul: I'd like some documentation about how effects are uploaded. For instance, the ID of the effect is only set in the last packet, how does the wheel know that the two preceding packets are supposed to apply to the third?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  memset(buf, 0, 15);
```
Kimplul: `memset` 15, but only 11 bytes are sent.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  /* Period (frequency) - default to 100ms = 10 Hz if not set */
+  if (period == 0) {
+    period = 100;
+  }
+
+  T500RS_DBG("Upload %s: id=%d, magnitude=%d (0x%02x), period=%dms\n",
+             type_name, effect->id, magnitude, mag, period);
+
+  /* Report 0x02 - Envelope */
+  memset(buf, 0, 15);
```
Kimplul: `memset` 15, but only 9 bytes sent.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  buf[14] = 0x00;
+  ret = t500rs_send_usb(t500rs, buf, 15);
+  if (ret) {
+    hid_err(t500rs->hdev, "Failed to send Report 0x01 (periodic main): %d\n",
+            ret);
+    return ret;
+  }
+
+  /* Report 0x04 - Periodic parameters */
+  memset(buf, 0, 15);
```
Kimplul: `memset` 15, but only 8 bytes sent.

src/hid-tmff2.c
```diff
@@ -743,6 +809,7 @@ static int tmff2_probe(struct hid_device *hdev, const struct hid_device_id *id)
 
 #if LINUX_VERSION_CODE < KERNEL_VERSION(6,12,0)
 static __u8 *tmff2_report_fixup(struct hid_device *hdev, __u8 *rdesc,
+
```
Kimplul: Weird place to put a newline?

Makefile
```diff
@@ -1,6 +1,17 @@
 KDIR ?= /lib/modules/$(shell uname -r)/build
 
+# Auto-generated build-time version for T500RS
```
Kimplul: This makefile is for all wheels, either make this be generic or just remove as I don't fully see the point of a miles long version string.

Out of curiosity, why do think it would be unfortunate to remove? Does this version string have any benefit that the DKMS version doesn't?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+    buf[2] = 0x00; /* STOP */
+    buf[3] = 0x01;
+    return t500rs_send_usb(t500rs, buf, 4);
+  }
+
+  /* For other effect types, send stop command - Report 0x41
+   * Use EffectID=0 to match device expectations for 0x41.
+   */
+  buf[0] = 0x41;
+  buf[1] = 0x00; /* Effect ID 0 */
```
Kimplul: Again, why does this only use effect ID 0?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+/* Update effect - re-upload and update force level if constant force */
+static int t500rs_update_effect(void *data,
+                                const struct tmff2_effect_state *state) {
+  struct t500rs_device_entry *t500rs = data;
+  const struct ff_effect *effect = &state->effect;
+
+  if (!t500rs)
+    return -ENODEV;
+
+  /* Do NOT re-upload here; Windows keeps the effect and only updates force
```
Kimplul: What? Not updating effects means that games can't modify effects they've already uploaded. Or am I missing something?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  buf[1] = 0x1c;
+  buf[2] = 0x00;
+  ret = t500rs_send_usb(t500rs, buf, 9);
+  if (ret) {
+    hid_err(t500rs->hdev, "Failed to send Report 0x02: %d\n", ret);
+    return ret;
+  }
+
+  /* Report 0x04 - Ramp parameters */
+  /* NOTE: T500RS doesn't support native ramp - just holds start level */
```
MmAaXx500: What does that mean? What is a "native" ramp? I would expect that the wheel handles the ramp effect, how did you determined that the ramp is not supported?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  if (ret)
+    return ret;
+
+  /* Report 0x05 - Condition parameters (deadband/center) */
+  memset(buf, 0, 15);
+  buf[0] = 0x05;
+  buf[1] = 0x1c;
+  buf[2] = 0x00;
+  buf[3] = 0x00; /* Deadband */
+  buf[4] = 0x00; /* Center */
```
MmAaXx500: Deadband and Center should be taken from the effect

MmAaXx500 (changes requested):
I have checked mostly from effects point of view.
Commended on some, but not all instance of the 0 effectids to do not spam too much.

Handling of envelopes are also missing/static in many if not all places (not commented on every instance).

There are also 99% static packets with not explained values, at least knowing why is it need would be good. (upload_constant, upload_condition, upload_periodic)

Replay `length` and `delay` is never taken into account (except `length` in ramp).

I think if would be good to document the packet formats, what is known and what is unknown in the packets, and what is the ordering of the packets, because many magic bytes are scattered all over the place.
T300 related effects are already documented in [docs/FFBEFFECTS.md](https://github.com/Kimplul/hid-tmff2/blob/master/docs/FFBEFFECTS.md) as an example.
Also using `struct`s instead of `buf[n]`s should be better as the code would be self documenting and it is a lot easier to catch errors.

Sorry for the many comments, I tried to not overlap with Kimplul's comments, but more I look the more falls out.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  return (transferred == len) ? 0 : -EIO;
+}
+
+/* Upload constant force effect */
+static int t500rs_upload_constant(struct t500rs_device_entry *t500rs,
+                                  const struct tmff2_effect_state *state) {
+  const struct ff_effect *effect = &state->effect;
+  u8 *buf = t500rs->send_buffer; /* Use DMA-safe buffer */
+  int ret;
+  int level = effect->u.constant.level;
```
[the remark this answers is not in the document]
Kimplul: Good catch! Seems to also apply to other effect uploads. Thanks for helping out, appreciate it.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  if (ret) {
+    hid_err(t500rs->hdev, "Failed to send Report 0x04: %d\n", ret);
+    return ret;
+  }
+
+  /* Report 0x01 - Main effect upload */
+  memset(buf, 0, 15);
+  buf[0] = 0x01;
+  buf[1] = effect->id;
+  buf[2] = 0x24; /* Ramp type (0x24 = sawtooth down / ramp) */
```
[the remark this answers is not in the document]
MmAaXx500: I think you are right, I didn't notice that and might explain the "time again?" comment, because the period of the periodic effect is exactly the same as the length of the effect to do not play the next period. That way you end up with a ramp. Awesome!

Can you update it?

cazzoo:
Very detailed review, thank you very much both for the feedback. Let me review the points and submit updated code

udev/99-thrustmaster.rules
tgurr: I've submitted https://github.com/berarma/oversteer/pull/309 to update the rule for oversteer. I was finally be able to test this PR today (successfully!) and tried to get rid of most of the udev rules like you suggested, together with the oversteer change breaking it down to what there is for other wheels just adding:

```
diff --git a/udev/71-thrustmaster-steamdeck.rules b/udev/71-thrustmaster-steamdeck.rules
index 3f3ea1e..ec15951 100644
--- a/udev/71-thrustmaster-steamdeck.rules
+++ b/udev/71-thrustmaster-steamdeck.rules
@@ -20,3 +20,6 @@ KERNEL=="hidraw*", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b692", MODE="0660
 
 # TSPC
 KERNEL=="hidraw*", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b689", MODE="0660", TAG+="uaccess"
+
+# T500RS
+KERNEL=="hidraw*", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", MODE="0660", TAG+="uaccess"
diff --git a/udev/99-thrustmaster.rules b/udev/99-thrustmaster.rules
index e96cb02..10c816e 100644
--- a/udev/99-thrustmaster.rules
+++ b/udev/99-thrustmaster.rules
@@ -27,3 +27,7 @@ SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b692", RUN+="/us
 # TSPC
 SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b689", RUN+="/usr/bin/evdev-joystick --evdev %E{DEVNAME} --deadzone 0"
 SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b689", RUN+="/usr/bin/jscal -s 6,1,1,32767,32767,16384,16384,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,0,0,0,536870912,536870912,1,0,0,0,536870912,536870912 /dev/input/js%n"
+
+# T500RS
+SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", RUN+="/usr/bin/evdev-joystick --evdev %E{DEVNAME} --deadzone 0"
+SUBSYSTEM=="input", ATTRS{idVendor}=="044f", ATTRS{idProduct}=="b65e", RUN+="/usr/bin/jscal -s 6,1,1,32767,32767,16384,16384,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,3,448,574,1394469,1394469,1,0,0,0,536870912,536870912,1,0,0,0,536870912,536870912 /dev/input/js%n"
```
things worked quite well for me, doing a quick test run in DiRT Rally (2015, has been delisted from steam in the meantime, running via Proton not the native version).

src/hid-tmff2.h
```diff
@@ -64,6 +64,12 @@ struct tmff2_device_entry {
 
 	spinlock_t lock;
 
+	/* Pending control changes to be applied from workqueue context */
```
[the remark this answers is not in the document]
cazzoo: While using usb interrupts, the workqueue deferral IS necessary because:

- The backend set_gain() and set_autocenter() functions call blocking USB operations (hid_hw_request(), usb_interrupt_msg())
- These cannot be called from atomic/interrupt context
- The FFB callbacks (  tmff2_set_gain,  tmff2_set_autocenter) can be called from atomic context
- Therefore, we must defer to a workqueue (process context)

We cannot use just flags without storing values because:

- For gain: The  value parameter from the FFB API is the application's requested gain, which is multiplied with the global  gain module parameter: (ffb_value * global_gain) / GAIN_MAX. We need to store the FFB value to apply this formula in the workqueue.
- For autocenter: The  value parameter from the FFB API needs to be stored to avoid race conditions (the callback could be called again before the workqueue processes the first request).

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+  size_t buffer_length;
+
+  /* Current wheel range for smooth transitions */
+  u16 current_range; /* Current rotation range in degrees */
+
+  /* Cache last-sent condition coefficients per effect ID to avoid re-sending
+   * identical 0x05 (sub=0x0e) coefficient updates at high cadence. ACC spams
+   * condition updates with identical values at low speed, and reapplying them
+   * causes tactile rumble on T500RS. Zero-initialized via kzalloc. */
+  u8 last_cond_rcoef[T500RS_MAX_EFFECTS];
```
Kimplul: This is new, but again duplicates data that is already available. `struct tmff2_effect_state` has both the new effect, `effect`, and the old effect that was uploaded, `old`, which can be used to do the diffing.

docs/FFBEFFECTS.md
```diff
@@ -1025,4 +1025,4 @@ binary    hex
     ff00.0021 = 0
     ff00.0021 = 0
     ff00.0021 = 0
-```
+```
```
Kimplul: Nitpick: whitespace change, try to avoid those. This seems to remove the final newline of the file.

docs/FFB_T500RS.md
```diff
@@ -0,0 +1,153 @@
+## T500RS USB FFB Protocol (observed)
```
Kimplul: I'd like some input from @MmAaXx500, but I personally find this documentation difficult to understand and I'd like to see a more example-driven document, where example packets are dissected, like in FFBEFFECTS.md

This documentation also doesn't seem to clarify *why* certain commands must use the effect slot `0x0`. Does the wheel only support a single effect at a time? Does `effect id` mean something else in these packets than it does in the T300?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+  if (ret) {
+    hid_err(t500rs->hdev, "Failed to send Report 0x03 (const level): %d\n",
+            ret);
+    return ret;
+  }
+
+  /* Report 0x01 - Main effect upload (second; references second envelope
+   * subtype) */
+  {
+    struct t500rs_r01_main *m = (struct t500rs_r01_main *)buf;
```
Kimplul: This is new. Why are you sending the same packet twice?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+  }
+
+  /* Report 0x03 - Constant force level (param subtype) */
+  {
+    s8 signed_level = t500rs_scale_const_level_s8(level);
+    struct t500rs_r03_const *r3 = (struct t500rs_r03_const *)buf;
+    r3->id = 0x03;
+    r3->code = 0x0e;
+    r3->zero = 0x00;
+    r3->level = signed_level;
```
Kimplul: Signed level is now sent to the wheel, but direction is still not taken into account.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+   */
+  /* Report 0x01 - Main effect upload - MATCH WINDOWS DRIVER EXACTLY! */
+  {
+    struct t500rs_r01_main *m = (struct t500rs_r01_main *)buf;
+    memset(m, 0, sizeof(*m));
+    m->id = 0x01;
+    m->effect_id = 0x00; /* Device expects Effect ID 0 for 0x01 on T500RS */
+    m->type = 0x00;      /* Constant force type */
+    m->b3 = 0x40;
+    m->b4 = 0xff; /* Windows uses 0xff (was 0x69) */
```
Kimplul: `b3` and `b4` are apparently the duration of the `FF_RAMP`. I would imagine that they're also the duration for other effects, so here you're setting the duration to be infinite, and then you rely on the driver to send a `STOP` command when the timer runs out, instead of letting the wheel take care of it. I think, not sure, needs better documentation.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+        if (idx < T500RS_MAX_EFFECTS && t500rs->last_cond_valid[idx] &&
+            t500rs->last_cond_rcoef[idx] == rcoef &&
+            t500rs->last_cond_lcoef[idx] == lcoef &&
+            t500rs->last_cond_b9[idx]    == b9    &&
+            t500rs->last_cond_b10[idx]   == b10) {
+          /* Coefficients unchanged: no-op (avoid rumble) */
+          return 0;
+        }
+
+        memset(buf, 0, 15);
```
Kimplul: Inconsistent buffer use, sometimes a packed struct is used, sometimes a raw buffer.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+  u8 *buf;
+  int ret;
+  u8 autocenter_percent;
+
+  if (!t500rs)
+    return -ENODEV;
+
+  autocenter_percent = (u8)((autocenter * 100) / 65535);
+
+  /* Wine compatibility: Some games (e.g., LFS under Wine) set autocenter to 100%%
```
Kimplul: Really? Which games? I don't remember seeing this with the T300.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+  struct t500rs_device_entry *t500rs = data;
+  u8 *buf;
+  int ret;
+  u16 range_value;
+
+  if (!t500rs)
+    return -ENODEV;
+
+  /* Clamp range to maximum value only
+   * Allow testing values below 270° to find hardware minimum */
```
Kimplul: Have you found the hardware minimum?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+  r->subtype = subtype;
+  r->zero = 0x00;
+  r->attack_len = cpu_to_le16(a_len);
+  r->attack_lvl = a_lvl;
+  r->fade_len = cpu_to_le16(f_len);
+  r->fade_lvl = f_lvl;
+}
+
+/* Debug logging helper (requires local variable named 't500rs') */
+#define T500RS_DBG(fmt, ...) hid_dbg(t500rs->hdev, fmt, ##__VA_ARGS__)
```
Kimplul: Still seems to use an implicit variable named `t500rs`.

Makefile
```diff
@@ -1,6 +1,17 @@
 KDIR ?= /lib/modules/$(shell uname -r)/build
 
+# Auto-generated global build-time version for TMFF2
+TMFF2_BASE_VERSION ?= 0.1
+GIT_HASH := $(shell git rev-parse --short=7 HEAD 2>/dev/null || echo "local")
```
Kimplul: @tgurr mentioned the implicit assumption that the module is built from a git repo, which might not hold.

src/tmt500rs/hid-tmt500rs-usb.c
cazzoo: Live For Speed, installed trough standard wine. You can give a try if you want, there a free demo version available over internet. It installs smoothly with bottles for instance (or legacy wine prefix)

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1502 @@
...
+  u8 b12;
+  u8 b13;
+  u8 b14;
+} __packed;
+
+/* Helper: classify whether a TX buffer is a known/managed report
+ * Known first bytes (report IDs / opcodes) we intentionally emit:
+ *   0x01,0x02,0x03,0x04,0x05,0x40,0x41,0x42,0x43,0x0a
+ */
+static inline int t500rs_is_known_tx(const unsigned char *data, size_t len) {
```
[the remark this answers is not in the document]
cazzoo: I swear I deleted it, once... I may have been lost in all my branches/rebases/merges.
I'll get rid of it for sure!

src/tmt500rs/hid-tmt500rs-usb.c
cazzoo: That is indeed a better way to handle it since this capability exists already, updating my code to use it!

docs/FFB_T500RS.md
cazzoo: I enriched documentation so it should be much more accurate, contains examples and sequences.
I took the opportunity of those documentation review // my captures to simplify the init sequence, that appears not to require the one I originally put in place.

cazzoo: That's a fair and valid point, so I re-wrote the FFB documentation for T500RS. I hope this is looking better now.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  /* Allocate send buffer */
+  t500rs->buffer_length = T500RS_BUFFER_LENGTH;
+  t500rs->send_buffer = kzalloc(t500rs->buffer_length, GFP_KERNEL);
+  if (!t500rs->send_buffer) {
+    ret = -ENOMEM;
+    goto err_buffer;
+  }
+
+  /* Initialize per-effect levels to 100% (0-100 range) */
+  t500rs->spring_gain = 100;
```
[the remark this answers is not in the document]
cazzoo: Resolved

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+
+  T500RS_DBG("Damper level set to %u%%\n", level);
+
+  /* Note: The new level will be applied when the next damper effect is
+   * uploaded/updated */
+  return 0;
+}
+
+/* Set friction level (0-100) - called from sysfs */
+static int t500rs_set_friction_level(void *data, u8 level) {
```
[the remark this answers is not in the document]
cazzoo: Removed

src/tmt500rs/hid-tmt500rs-usb.c
cazzoo: Effects are piling over the previous one as per my testing

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  default:
+    hid_err(t500rs->hdev, "Unknown periodic waveform: %d\n",
+            effect->u.periodic.waveform);
+    return -EINVAL;
+  }
+
+  /* Magnitude - scale to 0-127 */
+  mag = (abs(magnitude) * 127) / 32767;
+
+  /* Period (frequency) - default to 100ms = 10 Hz if not set */
```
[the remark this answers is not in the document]
Kimplul: This is still not resolved. Line number has changed, but the assumption about a default period exists in `t500rs_upload_periodic`.

src/tmt500rs/hid-tmt500rs-usb.c
Kimplul: So `effect_id` doesn't encode an effect identifier, since all effects have the same ID of 0...? Not sure I follow.

For instance, let's say a game uploads two effects, a constant force and a periodic force, and sets them both playing. How does the game now stop only the periodic effect?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  T500RS_DBG("Setting wheel range to %u degrees\n", range);
+
+  /* Based on testing with actual hardware:
+   * The T500RS uses Report 0x40 0x11 [value_lo] [value_hi] to set rotation
+   * range
+   *
+   * Hardware testing showed:
+   * - Byte order is LITTLE-ENDIAN (low byte first)
+   * - Formula: value = range * 60
+   * - Smooth transitions prevent hard mechanical ticking
```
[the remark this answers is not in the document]
Kimplul: I'm still curious, what does mechanical ticking mean in this context?

src/hid-tmff2.c
```diff
@@ -315,15 +376,26 @@ static void tmff2_work_handler(struct work_struct *w)
...
 		effect_delay = state->effect.replay.delay;
 		effect_length = state->effect.replay.length;
+		/* If playing with a finite length, stop when (delay + length) elapses */
 		if (test_bit(FF_EFFECT_PLAYING, &state->flags) && effect_length) {
 			if ((time_now - state->start_time) >=
 					(effect_delay + effect_length) * state->count) {
 				__clear_bit(FF_EFFECT_PLAYING, &state->flags);
 				__clear_bit(FF_EFFECT_QUEUE_UPDATE, &state->flags);
-
+				/* Request a STOP in process context */
```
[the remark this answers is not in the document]
Kimplul: > I empirically saw the T500 keep outputting force after Linux considered the effect finished

Yes, but that's because you don't seem to encode the duration of the effect in the packets. How would the wheel know when to stop the effect if you don't tell it how long to play the effect for?

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1260 @@
...
+  const struct ff_effect *effect = &state->effect;
+  u8 *buf = t500rs->send_buffer; /* Use DMA-safe buffer */
+  int ret;
+  int start_level = effect->u.ramp.start_level;
+  int end_level = effect->u.ramp.end_level;
+  u16 duration_ms = effect->replay.length;
+  u16 start_scaled;
+
+  /* Scale to 0-255 */
+  start_scaled = (abs(start_level) * 0xff) / 32767;
```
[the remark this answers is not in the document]
Kimplul: Seems like you take the direction of the effect into account for constant forces, but to be clear, that should also be done for periodic effects.

src/tmt500rs/hid-tmt500rs-usb.c
Kimplul: Alright, then please clamp the range to a minimum of 40.

src/tmt500rs/hid-tmt500rs-usb.c
```diff
@@ -0,0 +1,1379 @@
...
+  struct usb_device *usbdev;
+  struct usb_interface *usbif;
+
+  int ep_out; /* INTERRUPT OUT endpoint address */
+
+  u8 *send_buffer;
+  size_t buffer_length;
+
+  /* Current wheel range for smooth transitions */
+  u16 current_range; /* Current rotation range in degrees */
```
Kimplul: I guess I missed this, but this is again work that is duplicated, you can just use the global `range` variable.

src/tmt500rs/hid-tmt500rs-usb.c
cazzoo: In fact, I introduced that smooth transition initially to prevent kicks while changing rotation from oversteer, but in fact, it is not necessary.

MmAaXx500:
> Another round of changes introduced based on your comments @Kimplul and @MmAaXx500 .
> If you mind having another review.
> I tried to consider all the comments, but maybe I missed some. Please let me know

Sure, once [my previous review](https://github.com/Kimplul/hid-tmff2/pull/175#pullrequestreview-3433013848) comments are answered with code change, or text reply. It starts to get hard to track that what is still valid from the previous review if we stack unanswered reviews on each other.

MmAaXx500:
> I don't know how you're testing the driver, [...]

Yeah, @cazzoo if you disclose that how you are trying to determine the various aspects of the wheel we might be able to suggest some improvements to it.
For example, I found that using Direct Input on Windows and controlling one parameter of an effect at the time makes it easier to isolate that what parameter corresponds to what bytes in the packets. That way, you can also verify it by writing the same program for Linux and comparing the produced packets. Of course, you can compare it by feeling, it's not that accurate, but testing whether the effect plays for 10s or 30s is good.

PR #186, opened 2025-12-09
hid: tmff2: add Thrustmaster T500RS wheel base driver

cazzoo:
This driver has been built from scratch based the previous dirty driver (see https://github.com/Kimplul/hid-tmff2/pull/175) and on captures from ffbsdl tool ran in windows for all possible effects (through SDL2 library).


src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+   * hw_slots_in_use = bitmap tracking which hardware slots are occupied
+   *
+   * Benefits: Reduced complexity, better cache performance, easier debugging.
+   */
+  u16 hw_id_map[T500RS_MAX_EFFECTS]; /* logical -> hardware mapping */
+  DECLARE_BITMAP(hw_slots_in_use,
+                 T500RS_MAX_HW_EFFECTS); /* occupied slots bitmap */
+
+  /*
+   * Thread safety - Phase 5 Security and Robustness.
```
Kimplul: What are these phases?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+    p->fade_level = t500rs_scale_envelope_level(env->fade_level);
+  } else if (env && !allow_nonzero) {
+    /*
+     * User requested envelope but device doesn't support it.
+     * Log once to inform user, then send zeros.
+     */
+    pr_warn_once("t500rs: Envelope requested but not supported for this effect type\n");
+  }
+
+  p->reserved = 0x00;
```
Kimplul: Since the packet is zero-initialized anyway, this is redundant.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+      case FF_INERTIA:
+        saturation = T500RS_SAT_INERTIA;
+        break;
+      default:
+        saturation = T500RS_SAT_DAMPER;
+        break;
+      }
+      struct t500rs_pkt_r05_condition *p =
+          (struct t500rs_pkt_r05_condition *)buf;
+      t500rs_build_r05_condition(p, (u8)(param_sub & 0xff), saturation);
```
Kimplul: Nitpick: `& 0xff` is not necessary when casting to `u8`.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+
+  /* Scale parameters using new protocol-accurate helpers */
+  mag = t500rs_scale_periodic_magnitude(effect->u.periodic.magnitude);
+  phase = t500rs_scale_periodic_phase(effect->u.periodic.phase);
+  offset = t500rs_scale_periodic_offset(effect->u.periodic.offset);
+  direction_dev = t500rs_scale_direction(effect->direction);
+  duration_ms = effect->replay.length ? effect->replay.length : 0xffff;
+  delay_ms = effect->replay.delay;
+  period_ms = effect->u.periodic.period;
+  if (period_ms == 0)
```
Kimplul: I've brought this up a few times, but a default period length is not correct, as far as I'm aware. Just remove it or report an error or something instead.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+  /* The remaining initialization (0x05 spring zeroing and 0x41 STOP for
+   * autocenter ID 15) is handled below.
+   */
+
+  /* Report 0x05 - Set deadband and center */
+  memset(init_buf, 0, 11);
+  init_buf[0] = 0x05;
+  init_buf[1] = 0x1c;
+  init_buf[2] = 0x00;
+  init_buf[3] = 0x00;  /* Deadband = 0 */
```
Kimplul: Why does the wheel have slots for deadband and center (assuming this comment is accurate) if they must be 0?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+  T500RS_DBG(t500rs, "T500RS: Cleaning up\n");
+
+  /* Free resources in reverse order of allocation */
+  if (t500rs->send_buffer) {
+    kfree(t500rs->send_buffer);
+    t500rs->send_buffer = NULL;
+  }
+
+  /* Clear the tmff2 data pointer to prevent use-after-free */
+  if (t500rs->hdev && t500rs->hdev->driver_data) {
```
Kimplul: Redundant.

src/tmt500rs/t500rs_protocol.h
```diff
@@ -0,0 +1,222 @@
...
+/*
+ * Packet Sequence Templates
+ *
+ * These arrays define the packet sequences for different effect types.
+ * Used by the packet sequencing abstraction system.
+ */
+/* Sequence templates are now static in the implementation file */
+
+/* Supported parameters and effects */
+extern const unsigned long t500rs_params;
```
Kimplul: These should just be `static` in `hid-tmt500rs.c`.

src/hid-tmff2.c
```diff
@@ -242,8 +242,31 @@ static ssize_t gain_store(struct device *dev,
...
+	* - The device callback receives the product: (pg * gain) / GAIN_MAX.
+	*   See worker at tmff2->set_gain(... (pg * gain) / GAIN_MAX ).
+	* When the sysfs `gain` changes, we trigger a recompute by pushing
+	* pending_gain_value = GAIN_MAX here so the effective device gain becomes
+	* exactly the sysfs value (GAIN_MAX * gain / GAIN_MAX == gain) and future
+	* in-game set_gain calls continue to multiply in.
+	*
+	* References:
+	* - docs/FFBEFFECTS.md: section "FF_GAIN" shows a dedicated device gain path.
+	* - docs/FFB_T500RS.md: Report glossary mentions 0x43 (gain), i.e., device-side
```
Kimplul: `docs/FFB_T500RS.md` doesn't exist.

src/hid-tmff2.c
```diff
@@ -242,8 +242,31 @@ static ssize_t gain_store(struct device *dev,
...
+	*   gain separate from per-effect magnitudes; drivers should expose both levels.
+	*/
+	if (tmff2->set_gain) {
```
Kimplul: Nitpick: this might be easier to read if you do `if (!tmff2->set_gain) return count;` and then do everything else.

src/hid-tmff2.c
```diff
@@ -242,8 +242,31 @@ static ssize_t gain_store(struct device *dev,
...
+		unsigned long flags;
+		spin_lock_irqsave(&tmff2->lock, flags);
+		tmff2->pending_gain_value = GAIN_MAX;
+		tmff2->gain_pending = 1;
```
Kimplul: I mentioned this in the previous PR but I'd really prefer it if you created a new `FF_EFFECT_QUEUE_XXX` flag instead of a separate `gain_pending` variable. Same goes for  autocentering.

src/hid-tmff2.c
```diff
@@ -315,15 +376,26 @@ static void tmff2_work_handler(struct work_struct *w)
...
 		effect_length = state->effect.replay.length;
+		/* If playing with a finite length, stop when (delay + length) elapses */
 		if (test_bit(FF_EFFECT_PLAYING, &state->flags) && effect_length) {
 			if ((time_now - state->start_time) >=
 					(effect_delay + effect_length) * state->count) {
 				__clear_bit(FF_EFFECT_PLAYING, &state->flags);
 				__clear_bit(FF_EFFECT_QUEUE_UPDATE, &state->flags);
-
+				/* Request a STOP in process context */
+				__set_bit(FF_EFFECT_QUEUE_STOP, &actions);
```
Kimplul: Why is this necessary if the wheel has a `duration` field?

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,663 @@
...
+#### Sequential Effects (No Overlap)
+
+| Scenario | Description | Effect IDs Used | Packet Sequence |
+|----------|-------------|-----------------|-----------------|
+| **Constant → Sine** | Constant 1500ms, then Sine 1500ms | 0x0000, then 0x0021 | Effect 1 (ID=0x00): Upload→Envelope→Constant→START→[wait]→STOP<br>Effect 2 (ID=0x21): Upload→Envelope→Periodic→START→[wait]→STOP |
+| **Sine → Triangle** | Sine 3000ms, then Triangle 3000ms | 0x0000, then 0x0022 | Effect 1 (ID=0x00): Upload→Envelope→Periodic→START→[wait]→STOP<br>Effect 2 (ID=0x22): Upload→Envelope→Periodic→START→[wait]→STOP |
+
+**Sequential Effect Notes:**
+- Each effect gets full upload sequence with unique effect ID
+- Windows driver uses non-sequential IDs: 0x00, 0x21, 0x22, 0x40, 0x41
```
Kimplul: What are these IDs? effect IDs only go up to 15, don't they?

Kimplul (changes requested):
Thanks for continuing on this project, I was getting worried that we scared you off :)

I did a quick overview. The documentation in particular looks a lot better, having clear examples of what each packet does and how they build up to a full effect upload helped greatly. I haven't done any cross-referencing between the docs and implementation, as I still found quite a lot of smaller issues I'd like taken care of first. I'll do a more thorough review after that.

Not entirely sure what you mean by 'from scratch', for instance the comment explaining the `set_gain` rationale is identical between this and the previous PR.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+  }
+
+  /* Validate gain range */
+  if (gain > T500RS_GAIN_MAX) {
+    hid_err(t500rs->hdev, "Gain %u exceeds maximum %d\n", gain,
+            T500RS_GAIN_MAX);
+    return -EINVAL;
+  }
+
+  t500rs = data;
```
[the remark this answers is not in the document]
cazzoo: Good catch

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+  }
+
+  /* Initialize device structure */
+  t500rs->hdev = tmff2->hdev;
+  t500rs->input_dev = tmff2->input_dev;
+  t500rs->current_range = 900; /* Default range: 900° */
+
+  /* Allocate send buffer with bounds checking */
+  t500rs->buffer_length = T500RS_BUFFER_LENGTH;
+  if (t500rs->buffer_length == 0 || t500rs->buffer_length > 4096) {
```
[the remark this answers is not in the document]
cazzoo: Removed

src/tmt500rs/hid-tmt500rs.c
cazzoo: Indeed, duplicate and non-necessary code

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+  int ret;
+
+  /* Input validation */
+  if (!data) {
+    pr_err("t500rs_set_gain: NULL data pointer\n");
+    return -ENODEV;
+  }
+
+  /* Validate gain range */
+  if (gain > T500RS_GAIN_MAX) {
```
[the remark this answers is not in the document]
cazzoo: Alright, removed

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+static int t500rs_play_effect(void *data,
+                              const struct tmff2_effect_state *state) {
+  struct t500rs_device_entry *t500rs = data;
+  const struct ff_effect *effect;
+  u8 *buf;
+  int ret;
+  int hw_id;
+
+  /* Input validation */
+  if (!data) {
```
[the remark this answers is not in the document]
cazzoo: I reviewed most of the input validation and kept only the ones I considered most important

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+   *
+   * Windows driver always sends zeros for periodic/constant:
+   * 02 38 00 00 00 00 00 00 00
+   */
+  if (env && allow_nonzero) {
+    p->attack_len = cpu_to_le16(env->attack_length);
+    p->attack_level = t500rs_scale_envelope_level(env->attack_level);
+    p->fade_len = cpu_to_le16(env->fade_length);
+    p->fade_level = t500rs_scale_envelope_level(env->fade_level);
+  } else if (env && !allow_nonzero) {
```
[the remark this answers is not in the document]
cazzoo: Simplified

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+      break;
+    case T500RS_SEQ_PERIODIC_RAMP:
+      if (effect->type == FF_RAMP) {
+        struct t500rs_pkt_r04_periodic_ramp *p =
+            (struct t500rs_pkt_r04_periodic_ramp *)buf;
+        t500rs_build_r04_ramp(p, (u8)(param_sub & 0xff),
+                              effect->u.ramp.start_level,
+                              effect->u.ramp.end_level, effect->replay.length);
+      } else {
+        u8 mag = t500rs_scale_periodic_magnitude(effect->u.periodic.magnitude);
```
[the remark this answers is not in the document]
cazzoo: Good catch, I re-implemented logic done in T300RS and it appears to be fine

src/tmt500rs/t500rs_protocol.h
```diff
@@ -0,0 +1,222 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
```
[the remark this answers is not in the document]
cazzoo: Renamed! that's indeed more clean

src/hid-tmff2.c
cazzoo: Good point, updated

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+    return -ENODEV;
+
+  autocenter_percent = (u8)((autocenter * 100) / 65535);
+
+  /* Wine compatibility: Some games (e.g., LFS under Wine) set autocenter to
+   * 100%% at startup and never release it. That leaves a permanent strong
+   * centering force which masks/overpowers other forces. To avoid this, ignore
+   * requests that try to set maximum autocenter (100%). Disabling (0) is still
+   * honored; lower values are allowed. */
+  if (autocenter_percent >= 100) {
```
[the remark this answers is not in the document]
cazzoo: Just double checked and reproduced the issue with RBR (RSF version). I don't know yet if that is a Wine bug or else, but this is really odd

src/hid-tmff2.c
cazzoo: From old doc file, sorted!

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+  ret = t500rs_send_hid(t500rs, init_buf, 11);
+  if (ret) {
+    hid_warn(t500rs->hdev, "Disable autocenter (0x05 0x1c) failed: %d\n", ret);
+  }
+
+  /* Stop autocenter effect (effect ID 15) */
+  {
+    struct t500rs_r41_cmd *r41 = (struct t500rs_r41_cmd *)init_buf;
+    r41->id = 0x41;
+    r41->effect_id = 15; /* Autocenter effect ID */
```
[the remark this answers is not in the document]
cazzoo: That was actually a good catch. I initially had it enabled at the very beginning (previous driver) and thought it was really necessary to get the forces working but I figured it is still working very fine without it, so removing from the init sequence

Makefile
```diff
@@ -1,6 +1,30 @@
...
 
+# Auto-generated global build-time version for TMFF2
+TMFF2_BASE_VERSION ?= 0.1
+
+# Allow packagers / CI to provide a fixed hash or full version:
+#   make GIT_HASH=deadbee
+#   make TMFF2_VERSION=0.1-1
+#
+# Only derive GIT_HASH from git if none was provided and this is a git checkout.
+ifeq ($(origin GIT_HASH), undefined)
```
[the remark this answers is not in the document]
cazzoo: I found it really useful while working on implementing the driver, making sure the loaded driver is the latest generated. I understand this may be over-complex and doesn't make full sense in several cases, I can definitely remove it if you don't see any benefits

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,663 @@
...
+| 0x49 | Polling | 7-16 bytes | Status polling (high frequency) |
+| 0x07 | Telemetry | 15 bytes | Position feedback (high frequency) |
+
+**Note:** Envelope support (0x02) is limited on T500RS hardware. Non-zero envelope values cause EPROTO errors on periodic and constant effects. Always send zeros for envelope parameters on these effect types.
+
+---
+
+## Subtype System and Effect Indexing
+
+On this wheel the last six bytes of the 0x01 main upload (bytes 9–14) do **not** contain envelope timings/levels directly. Instead they carry two 16‑bit "subtype" values that act as per‑effect indices:
```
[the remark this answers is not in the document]
cazzoo: Correct, I'd add that notion

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,663 @@
...
+- **Examples:**
+  - SDL 0 → Device 0
+  - SDL 8000 → Device 6
+  - SDL 16000 → Device 12
+  - SDL 24000 → Device 18
+  - SDL 32767 → Device 255
+
+---
+
+## Important implementation facts
```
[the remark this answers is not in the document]
cazzoo: Fair, just removed and kept locally for my personal docs

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+/* Supported parameters */
+const unsigned long t500rs_params = PARAM_SPRING_LEVEL | PARAM_DAMPER_LEVEL |
+                                    PARAM_FRICTION_LEVEL | PARAM_GAIN |
+                                    PARAM_RANGE;
+
+/*
+ * Supported effects.
+ *
+ * NOTE: FF_SQUARE is intentionally OMITTED. Per Windows USB captures, the
+ * T500RS protocol does not encode waveform type in USB packets. Windows/SDL2
```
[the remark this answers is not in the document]
cazzoo: I managed to get SQUARE wave implemented, with some updated code from ffbsdl and fedit.exe. Will update the PR

Makefile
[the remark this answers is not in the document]
cazzoo: Now I have a pretty decent codebase, I reverted because this is indeed not particularly necessary.

cazzoo:
Hi @Kimplul ,

I've been working again on the PR, responding to all your points (almost all -- if not all -- were relevant), and introducing some missing features (FF_SQUARE missing and 0x05 conditional effect that was incorrectly done). If you mind having another pass, I'd be really happy. 
Thanks

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+        envelope = &effect->u.ramp.envelope;
+        allow_envelope = true;  /* Ramp supports envelope */
+      } else if (effect->type == FF_CONSTANT || effect->type == FF_PERIODIC) {
+        envelope = &effect->u.periodic.envelope;
+        allow_envelope = false;  /* Firmware bug: must send zeros */
+      }
+
+      t500rs_build_r02_envelope(env, (u8)(env_sub & 0xff), envelope, allow_envelope);
+    }
+      ret =
```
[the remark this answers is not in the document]
Kimplul: Looks like you missed a few, please check `t500rs_send_packet_sequence` starting with `T500RS_SEQ_ENVELOPE`.

src/tmt500rs/hid-tmt500rs.h
```diff
@@ -0,0 +1,224 @@
...
+#define T500RS_EFFECT_INERTIA 0x41
+
+/* Saturation values for conditional effects */
+#define T500RS_SAT_SPRING 84
+#define T500RS_SAT_DAMPER 100
+#define T500RS_SAT_FRICTION 100
+#define T500RS_SAT_INERTIA 100
+
+/* Hardware limits */
+#define T500RS_MAX_EFFECTS 16
```
Kimplul: Now that you're using the ID mapping of `logical_id + 1`, the wheel can only handle 15 effects simultaneously, I think?

src/tmt500rs/hid-tmt500rs.h
```diff
@@ -0,0 +1,224 @@
...
+  u8 id;             /* 0x02 */
+  u8 subtype;        /* from 0x01 code2 (env_sub low byte) */
+  __le16 attack_len; /* attack duration in ms */
+  u8 attack_level;   /* 0-255 */
+  __le16 fade_len;   /* fade duration in ms */
+  u8 fade_level;     /* 0-255 */
+  u8 reserved;       /* 0x00 */
+} __packed;
+
+#endif /* __T500RS_PROTOCOL_H */
```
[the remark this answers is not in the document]
cazzoo: Got it fixed, typo while renaming

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,2073 @@
...
+
+/*
+ * Scale direction from Linux ff_effect format to T500RS protocol format.
+ *
+ * Linux ff_effect.direction: 0-65535 (0 = forward, 16384 = right, 32768 = back,
+ * 49152 = left) T500RS protocol: 0-35999 in 0.01 degree units (0 = 0°, 9000 =
+ * 90°, 18000 = 180°, etc.)
+ *
+ * Conversion: device_dir = (linux_dir * 36000) / 65536
+ * This maps 0-65535 → 0-35999 (approximately, since 65535 → 35999.45)
```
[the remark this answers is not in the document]
cazzoo: Good catch, fixed

src/tmt500rs/hid-tmt500rs.h
cazzoo: Well, The absolute number of effects the wheel supports is 16, constant included. I believe constant can be considered as an effect, and since the logic is doing all the magic to prevent using slot 0 I'd assume we can keep this way.

I you believe we prefer having "MAX_DYNAMIC_EFFECTS" alike variable (set to 15), I'm fine too, it may simplify the logic here or there, but it's literally something we should agree together. Whatever is fine to me, just let me know what suits you the most

cazzoo:
No worries, thank you for the feedback you provided. Take the necessary time for whatever need review. I am not in the hurry, and this driver does work for me already so It can last as long as necessary from your end.

I'm happy we trend to get somewhat a stable mergeable version.
I just pushed a new part of code that handled previously the mode switch (from boot/init to normal mode). This is the part where I have the least confidence. The logic is fully fine, it's working well, but I'm not sure at all on code quality and the compliance/integration with the base driver. If you may give a glance when you can, I'd be happy to send some more time if needed.

Cheers

src/tmt500rs/hid-tmt500rs.h
Kimplul: The main thing is just that `input_ff_create` in https://github.com/Kimplul/hid-tmff2/blob/26995429a3b92047f0c2f9ab90fe4202c86fd72b/src/hid-tmff2.c#L627 needs to know how many effects can be uploaded to the wheel at a time. I don't think I've seen 16 effects active at the same time, but if such a situation occurs, we want the FFB subsystem to report to the application that it's not supported, rather than send possibly-garbage packets to the wheel.

Kimplul:
> EDIT: I just open a PR against hid-tminit for TSCP branch (even though it's t500rs code update)

Yeah, I intended to merge `tspc` into `usb` as part of adding support for the TS-PC, but seems I forgot. I'll sort it out once we deal with this PR, don't want to start changing too many moving bits at the same time.

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,574 @@
...
+
+**Common Code Combinations:**
+- Constant effects: bytes 9-10 = 0x000e (for 0x03 packet), bytes 11-12 = 0x001c (envelope)
+- Periodic effects: bytes 9-10 = 0x002a (for 0x04 packet), bytes 11-12 = 0x001c (envelope)
+- Conditional effects: bytes 9-10 = 0x002a (for first 0x05 packet), bytes 11-12 = 0x0038 (for second 0x05 packet)
+- Alternative codes observed: 0x00b6/0x00c4 (newer captures), 0x0046/0x0054, 0x0062/0x0070, 0x007e/0x008c, 0x009a/0x00a8
+
+**Examples:**
+- `01 00 00 40 f4 01 00 00 0e 00 1c 00 00 00` - Constant effect with envelope
+  - Effect ID: 0x00
```
Kimplul: Examples should probably not use ID 0, since the driver doesn't use it.

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,574 @@
...
+- Magnitude (0-32767) → Device magnitude (0-127)
+- Phase (0-35999, 0.01° units) → Device phase (0-255, 256 steps for 360°)
+- Period in milliseconds (no conversion needed)
+- Code 0x2a is used (NOT 0x0e as in current driver!)
+
+### 3. PERIODIC EFFECTS - TRIANGLE WAVE
+
+| Test Case | SDL2 Parameters | Packet Sequence | Packet Payloads |
+|-----------|----------------|-----------------|-----------------|
+| **Triangle Medium** | mag=24000, period=100ms, phase=180°, dir=180° | 1. Upload<br>2. Envelope<br>3. Periodic<br>4. START | `01 00 00 40 d0 07 00 00 0e 00 1c 00 00 00`<br>`02 1c 00 00 12 00 00 12 00`<br>`04 2a 09 00 7f 64 00 00`<br>`41 00 41 01` |
```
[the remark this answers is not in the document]
Kimplul: While I'm at it, it would maybe be nice if each example not only showed what bytes are sent, but also how they are encoded, like in the docs for the T300. Like

```
0x10 - Effect ID 
0x20 - Effect Code
```

and so on, the current examples require a lot of going back and forth between the different packet formats to figure out what each byte actually means.

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,574 @@
...
+| Test Case | SDL2 Parameters | Packet Sequence | Packet Payloads |
+|-----------|----------------|-----------------|-----------------|
+| **Damper Low** | right_coeff=low, left_coeff=low, right_sat=max, left_sat=max | 1. Upload<br>2. Cond Axis 1<br>3. Cond Axis 2<br>4. START | `01 00 40 40 d0 07 00 00 2a 00 38 00 00 00`<br>`05 2a 00 00 00 00 00 00 00 64 64`<br>`05 38 00 00 00 00 00 00 00 64 64`<br>`41 00 41 01` |
+| **Damper High** | right_coeff=high, left_coeff=high | 1. Upload<br>2. Cond Axis 1<br>3. Cond Axis 2<br>4. START | `01 00 40 40 d0 07 00 00 2a 00 38 00 00 00`<br>`05 2a 00 00 00 00 00 00 00 64 64`<br>`05 38 00 00 00 00 00 00 00 64 64`<br>`41 00 41 01` |
+
+**Note:** Damper uses same 0x05 packet structure as spring. Saturation values differ (0x6464 for damper vs 0x5454 for spring).
+
+### 10. CONDITIONAL EFFECTS - INERTIA
+
+**Capture Example:**
```
Kimplul: This is a better way to handle examples of each effect upload type, why did the style change midway through the document?

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,574 @@
...
+- **Conversion:** Scale down
+- **Formula:** `device_env = sdl_env * 255 / 32767`
+- **Examples:**
+  - SDL 0 → Device 0
+  - SDL 8000 → Device 6
+  - SDL 16000 → Device 12
+  - SDL 24000 → Device 18
+  - SDL 32767 → Device 255
+
+---
```
Kimplul: Stuff after this line seems to be irrelevant, please remove.

src/tmt500rs/hid-tmt500rs.c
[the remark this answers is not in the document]
Kimplul: I think this should be removed, as I don't think it's the driver's responsibility to fix broken games. If the autocentering is too strong, it can be turned down in Oversteer.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1622 @@
...
+    hid_warn(t500rs->hdev,
+             "Ignoring 100%% autocenter request (Wine/LFS compatibility)");
+    return 0;
+  }
+
+  buf = t500rs->send_buffer;
+  if (!buf)
+    return -ENOMEM;
+
+  /* Enable autocenter: Report 0x40 0x04 0x01 */
```
Kimplul: Nitpick: structures could be used here as well to give each field a name at least.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1622 @@
...
+      break;
+    }
+    struct t500rs_pkt_r05_condition *p = (struct t500rs_pkt_r05_condition *)buf;
+    t500rs_build_r05_condition(p, (u8)(param_sub & 0xff), cond->right_coeff,
+                               cond->left_coeff, saturation, cond->deadband,
+                               cond->center);
+    return t500rs_send_hid(t500rs, buf, sizeof(*p));
+  }
+  default:
+    return 0;
```
Kimplul: Shouldn't this be an error since the effect type is either not supported or corrupted?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1622 @@
...
+  }
+
+  hw_id = t500rs_logical_to_hw_id(state->effect.id);
+
+  /* STOP command uses hw_id as effect_id to match 0x01 packet */
+  return t500rs_send_stop(t500rs, (u8)hw_id);
+}
+
+/* Update effect - send parameter updates without re-uploading */
+static int t500rs_update_effect(void *data,
```
Kimplul: Effect updating doesn't seem to handle effect durations/delays or other things that would be part of the `0x01` packet type. Are those unsupported by the wheel or left out to simplify stuff? I think we had a discussion about this but I can't seem to find it at the moment. Either way, that should be documented somewhere. I don't think I've seen modifying effect durations or delays in the wild so it's probably fine, but something to keep in mind.

Kimplul (changes requested):
Sorry about the delay, I've been busier than expected. I mainly focused on the docs and found some inaccuracies and points of improvement, please have a look at those.

I applied some style fixes myself, please cherry-pick them from https://github.com/Kimplul/hid-tmff2/tree/cazzoo-hardening/t500rs-driver. I don't have push access to your branch and opening a PR for a PR seemed a bit silly, but the changes were trivial enough that I didn't really want to request them explicitly. Kind of a clumsy part of the PR flow, but eh.

Regarding style, I've tried to follow the Linux Kernel style guide in this repo. Most significantly, it dictates a tab width of 8 spaces, whereas this has a width of 2, please fix that. You might also want to try running your code through `clang-format` with the kernel's style configuration, see https://www.kernel.org/doc/html/next/dev-tools/clang-format.html

src/tmt500rs/hid-tmt500rs.c
cazzoo: With T500RS I observed that packets are uploaded once with duration, and then the subsequent updates provides parameter packets that contain forces parameters. If we would like to see duration/delays updated, it would mean a new 0x01 packet to be sent, with a cancellation packet before, and I haven't observed it in my captures. Maybe I missed something but I haven't with all my attempts. Probably real captures from windows would catch this behavior but I don't have and cannot collect by my own anymore.
So, in order to answer you point, I'll add the rationale of this in the code and documentation, just so it becomes explicit.

docs/T500RS_USB_Protocol_Analysis.md
cazzoo: That's historical, but you're right, so I tided and unified the doc format/structure

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,574 @@
...
+| **Sine Low** | mag=8000, period=10ms, phase=90°, dir=90° | 1. Upload<br>2. Envelope<br>3. Periodic<br>4. START | `01 00 00 40 d0 07 00 00 0e 00 1c 00 00 00`<br>`02 1c 00 00 06 00 00 06 00`<br>`04 2a 06 00 3f 0a 00 00`<br>`41 00 41 01` |
+| **Sine Medium** | mag=24000, period=100ms, phase=180°, dir=180° | 1. Upload<br>2. Envelope<br>3. Periodic<br>4. START | `01 00 00 40 d0 07 00 00 0e 00 1c 00 00 00`<br>`02 1c 00 00 12 00 00 12 00`<br>`04 2a 09 00 7f 64 00 00`<br>`41 00 41 01` |
+| **Sine with Envelope** | mag=24000, period=100ms, attack=500ms, fade=500ms | 1. Upload<br>2. Envelope<br>3. Periodic<br>4. START | `01 00 00 40 d0 07 00 00 0e 00 1c 00 00 00`<br>`02 1c f4 01 12 f4 01 12 00`<br>`04 2a 09 00 00 64 00 00`<br>`41 00 41 01` |
+
+**Periodic Effect Notes:**
+- Magnitude (0-32767) → Device magnitude (0-127)
+- Phase (0-35999, 0.01° units) → Device phase (0-255, 256 steps for 360°)
+- Period in milliseconds (no conversion needed)
+- Code 0x2a is used (NOT 0x0e as in current driver!)
```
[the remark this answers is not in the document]
cazzoo: True, there were some inconsistencies, doc updated

docs/T500RS_USB_Protocol_Analysis.md
```diff
@@ -0,0 +1,574 @@
...
+
+---
+
+## Effect Type Implementation Table
+
+### 1. CONSTANT FORCE EFFECTS
+
+| Test Case | SDL2 Parameters | Packet Sequence | Packet Payloads |
+|-----------|----------------|-----------------|-----------------|
+| **Constant Zero** | level=0, dir=0°, len=500ms | 1. Upload<br>2. Envelope<br>3. Constant<br>4. START | `01 00 00 40 f4 01 00 00 0e 00 1c 00 00 00`<br>`02 1c 00 00 00 00 00 00 00`<br>`03 0e 00 00`<br>`41 00 41 01` |
```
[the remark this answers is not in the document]
cazzoo: I re-wrote the template as per one of the examples that were fine later in the doc, it should be better now I believe

docs/T500RS_USB_Protocol_Analysis.md
cazzoo: Corrected

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+The T500RS is a single-axis force feedback wheel with a rotating range that can be configured (typically 900 degrees or 1080 degrees). It uses a proprietary USB protocol for force feedback effects, distinct from the T300RS and other Thrustmaster wheels.
+
+### Effect Playing and Stopping
+
+#### Playing an Effect
+```
+41 00 41 01 - START command
+```
+- **Packet Type:** 0x41 (Command)
+- **Effect ID:** 0x00 (always 0x00 for T500RS)
```
Kimplul: I believe we've already established that the effect ID varies depending on what effects are loaded, no?

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+- **Command:** 0x41 (START)
+- **Argument:** 0x01 (play count, 0x01 = play once)
+
+#### Stopping an Effect
+```
+41 00 00 01 - STOP command
+```
+- **Packet Type:** 0x41 (Command)
+- **Effect ID:** 0x00 (always 0x00 for T500RS)
```
Kimplul: Ditto.

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+Offset | Size | Field          | Description
+-------|------|----------------|----------------------------------
+ 0      | 1    | packet_type    | 0x01
+ 1      | 1    | effect_id      | Hardware effect slot ID (0-15, assigned by driver)
+ 2      | 1    | effect_type    | Effect type (see table below)
+ 3      | 1    | control        | Always 0x40
+ 4      | 2    | duration_ms    | Duration in milliseconds, little-endian
+ 6      | 2    | delay_ms       | Delay before start, little-endian
+ 8      | 1    | reserved1      | 0x00
+ 9      | 2    | packet_code_1  | Code for subsequent packet type (variable!)
```
Kimplul: `packet_code_1` and `packet_code_2` are the same as `param_sub` and `envelope_sub`, are they not? I think `param_sub` and `envelope_sub` are more descriptive terms, the code uses them as well.

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+| 0x23 | Sawtooth Up | Inferred from pattern |
+| 0x24 | Sawtooth Down | Inferred from pattern |
+| 0x40 | Spring | Windows driver captures |
+| 0x41 | Damper/Friction/Inertia | Windows driver + FFEdit captures |
+
+**Note:** Square wave (0x20) was discovered in FFEdit captures. The Windows driver may not expose this effect type through the standard API.
+
+**IMPORTANT:** Bytes 9-12 specify the packet codes used in subsequent packets. These are NOT fixed values!
+
+**Common Code Combinations:**
```
Kimplul: Again, without explaining subtype indexing this information is not useful to someone reading throught this document for the first time. It's extra noise that makes it more difficult to understand the whole picture.

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+
+**Examples:**
+- `01 01 00 40 f4 01 00 00 0e 00 1c 00 00 00` - Constant effect with envelope
+  - Effect ID: 0x01 (hardware slot 1, logical 0)
+  - Effect type: 0x00 (constant)
+  - Control: 0x40
+  - Duration: 0x01f4 = 500ms
+  - Delay: 0x0000 = 0ms
+  - Reserved1: 0x00
+  - Packet codes: 0x000e (constant), 0x001c (envelope)
```
Kimplul: I guess it would make more sense to describe how each `param_sub`, `envelope_sub` is calculated (`some_constant + some_other_constant * effect_id`)

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+2      | 1    | magnitude      | 0-127 (effect strength)
+3      | 1    | offset         | Signed -127 to +127 (DC offset)
+4      | 1    | phase          | 0-255 (0-360 degrees, 256 steps)
+5      | 2    | period_ms      | Period in milliseconds, little-endian
+7      | 1    | reserved       | 0x00
+```
+
+**Code Values:** The code in byte 1 is the low byte of the parameter_subtype from the 0x01 packet (dynamically calculated based on effect index)
+
+**Period Encoding:** Period is in MILLISECONDS (not Hz*100). No conversion needed.
```
Kimplul: Here again, `(not Hz*100)` doesn't seem like important information for a first time reader. Just say the period is in milliseconds, this just makes me want to ask "why is Hz*100 brought up?" which detracts from getting a full picture of what's going on.

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+- `41 00 41 01` - START effect
+- `41 00 00 01` - STOP effect
+
+---
+
+## Effect Type Implementation Table
+
+### 1. CONSTANT FORCE EFFECTS
+
+**Capture Examples:**
```
Kimplul: These examples don't seem to even try to explain what's going on. You don't necessarily have to go through each packet field by field, but at least tell the reader what each packet is/does. "Set constant force of zero for 100ms for effect slot 2: Send packet 0x01 with duration 100ms: <packet> Set constant force strength: <packet>` or something along those lines.

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+  - effect_type = 0x22 (sine wave)
+  - codes: 0x002a (periodic parameters), 0x001c (envelope)
+- Envelope packet: `02 1c [attack_len] [attack_lvl] [fade_len] [fade_lvl] 00`
+- Periodic packet: `04 2a [magnitude] [offset] [phase] [period_ms] 00`
+
+**Parameter Details:**
+- Magnitude: 0-127 (scaled from Linux 0-32767)
+- Offset: s8 (-128 to +127, DC bias (Direct Current bias - a constant force offset), scaled from Linux -32768 to +32767)
+- Phase: 0-255 (256 steps for 360 degrees, scaled from Linux 0-35999)
+- Period: Direct milliseconds (no Hz conversion!)
```
Kimplul: What is this `Hz conversion`?

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+### 1. CONSTANT FORCE EFFECTS
+
+**Capture Examples:**
+- Zero force: `01 00 00 40 f4 01 00 00 0e 00 1c 00 00 00` `02 1c 00 00 00 00 00 00 00` `03 0e 00 00`
+- Low positive force: `01 00 00 40 d0 07 00 00 0e 00 1c 00 00 00` `02 1c 00 00 06 00 00 06 00` `03 0e 00 03`
+- Medium force: `01 00 00 40 d0 07 00 00 0e 00 1c 00 00 00` `02 1c 00 00 12 00 00 12 00` `03 0e 00 09`
+- High negative force: `01 00 00 40 88 13 00 00 0e 00 1c 00 00 00` `02 1c 00 00 0d 00 00 0d 00` `03 0e 00 f9`
+- Maximum force with direction: `01 00 00 40 d0 07 00 00 0e 00 1c 00 00 00` `02 1c 00 00 00 00 00 00 00` `03 0e 00 00`
+
+**Packet Structure:**
```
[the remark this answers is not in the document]
Kimplul: Also, present the format first, then examples.

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+|-----------------|-------------------|------------------|
+| 1               | 0x002a            | 0x0038           |
+| 2               | 0x0046            | 0x0054           |
+| 3               | 0x0062            | 0x0070           |
+| 4               | 0x007e            | 0x008c           |
+| 5               | 0x009a            | 0x00a8           |
+| 6               | 0x00b6            | 0x00c4           |
+
+### Driver Implementation Notes
+- **Effect ID Handling:** The driver uses hardware IDs 1-15 to avoid quirky behavior with hardware index 0 (only valid for constant effects).
```
Kimplul: This is already mentioned above. Sometimes repetition can be useful, but there's only like 15 lines between each occurence so one or the other should probably be removed.

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,564 @@
...
+---
+
+## Parameter Encoding Reference
+
+### Direction Encoding
+- **Linux FFB Format:** 0-65535 (0 = forward, 16384 = right, 32768 = back, 49152 = left)
+- **Device Format:** 16-bit little-endian (0-35999 in 0.01 degree units)
+- **Conversion:** `device_dir = (os_ffb_dir * 36000) / 65536`
+- **Examples:**
+  - 0degrees = 0x0000
```
Kimplul: I believe it's customary to put a space between the number and `degrees`, presumably because `degree` is a whole word. Short forms, such as `ms` for millisecond, can be written without a space between.

Kimplul (commented):
At least the documentation still needs a bit of work. I would suggest maybe to read through everything a couple of times from a perspective of someone who doesn't know anything about the wheel, reading from top to bottom to get a good overall picture of what the wheel does and how. Present things in the order that they're needed, lots of examples, try to maintain a consistent way to present information, stuff like that.

SeitzB:
I have been following this development with some excitement and just tried to build and run the current PR. I do not know how to properly mention files in a comment, but in the hid-tmt500rs-usb.c file there is a capital i after a semicolon.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+
+	/* Project magnitude based on direction (same formula as T300RS) */
+	projected =
+		(os_ffb_mag * fixp_sin16(direction * 360 / 0x10000)) / 0x7fff;
+
+	if (projected < 0) {
+		/* Wheel handles positive magnitudes only */
+		projected = -projected;
+
+		/* Add 180 degrees to phase to maintain correct force direction.
```
Kimplul: Same here

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+	/* Compute average magnitude from start/end levels */
+	avg_level = (abs(start_level) + abs(end_level)) / 2;
+	magnitude = (u8)((avg_level * 127) / 32767);
+
+	/* Offset encodes direction: positive = ramping up, negative = ramping down */
+	/* Simple approximation: (end - start) / 512 to fit in s8 range */
+	offset = (s8)((end_level - start_level) / 512);
+
+	/*
+   * Phase encodes ramp direction per FFEdit captures:
```
Kimplul: Same

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+ */
+static void t500rs_build_r02_envelope(struct t500rs_pkt_r02_envelope *p,
+				      u8 subtype, const struct ff_envelope *env,
+				      bool allow_nonzero)
+{
+	memset(p, 0, sizeof(*p));
+	p->id = 0x02;
+	p->subtype = subtype;
+
+	/*
```
Kimplul: Same

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+   * 02 38 00 00 00 00 00 00 00
+   */
+	if (env && allow_nonzero) {
+		p->attack_len = cpu_to_le16(env->attack_length);
+		p->attack_level =
+			t500rs_scale_envelope_level(env->attack_level);
+		p->fade_len = cpu_to_le16(env->fade_length);
+		p->fade_level = t500rs_scale_envelope_level(env->fade_level);
+	} else {
+		/*
```
Kimplul: Same

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+			bool allow_envelope = false;
+
+			if (effect->type == FF_RAMP) {
+				envelope = &effect->u.ramp.envelope;
+				allow_envelope =
+					true; /* Ramp supports envelope */
+			} else if (effect->type == FF_CONSTANT ||
+				   effect->type == FF_PERIODIC) {
+				envelope = &effect->u.periodic.envelope;
+				allow_envelope =
```
Kimplul: This just looks funny :p

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+				sizeof(struct t500rs_pkt_r04_periodic_ramp));
+			if (ret)
+				break;
+			break;
+		}
+
+		case T500RS_SEQ_CONDITION_X: {
+			const struct ff_condition_effect *cond =
+				&effect->u.condition[0];
+			/* Scale saturation from Linux FFB range (0..65535) to device range
```
Kimplul: Comment spacing issue continues here

Kimplul:
@SeitzB I think you have the wrong branch checked out, `master` seems to have the spurious `I` while the most recent changes are in `hardening/t500rs-driver`.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+			"it back to expected value using oversteer (or keep oversteer "
+			"open) or system gain.");
+	}
+
+	buf = t500rs->send_buffer;
+	if (!buf)
+		return -ENOMEM;
+
+	/* Enable autocenter: Report 0x40 0x04 0x01 */
+	{
```
Kimplul: This extra block is unnecessary

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+
+	/* Store device data in tmff2 BEFORE any operations that might fail */
+	tmff2->data = t500rs;
+
+	/* Use send_buffer for all HID transfers */
+	init_buf = t500rs->send_buffer;
+
+	T500RS_DBG(t500rs, "Sending initialization sequence...\n");
+
+	/* Report 0x42 - Init/status commands (2 bytes each)
```
Kimplul: More comment whitespace issue

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+		// Keep explicit zeros even though memset() clears them.
+		config->data1 = 0x00;
+		config->data2 = 0x00;
+	}
+	ret = t500rs_send_hid(t500rs, init_buf, 4);
+	if (ret)
+		hid_warn(t500rs->hdev,
+			 "Init command 3 (0x40 config) failed: %d\n", ret);
+
+	/* Report 0x43 - Set global gain (2 bytes)
```
Kimplul: Same

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1765 @@
...
+						   cond->left_coeff, right_sat,
+						   left_sat, cond->deadband,
+						   cond->center);
+			ret = t500rs_send_hid(
+				t500rs, buf,
+				sizeof(struct t500rs_pkt_r05_condition));
+			break;
+		}
+
+		case T500RS_SEQ_CONDITION_Y: {
```
Kimplul: Does anyone use the second axis? The T300 only uses `effect->u.condition[0]`, and I don't remember seeing kernel drivers use `condition[1]`.

Kimplul (commented):
I added some comments about the most recent stylistic changes. Looks like `clang-format` did something weird with the whitespacing in multi-line comments, or maybe they were weird to begin with and `clang-format` just choked. Dunno.

Kimplul:
@hoover67 This should work:

```
git clone --recurse-submodules --branch=hardening/t500rs-driver https://github.com/cazzoo/hid-tmff2.git
```

From there on, you should be able to just follow the installation instructions in the README.

hoover67:
Thanks a lot for these instructions! 

Where should I report errors during compilation? I tried both methods (make and dkms) and both fail with the same compiler error message: 

```
 CC [M]  /home/hoover/src/hid-tmff2/src/tmtspc/hid-tmtspc.o
  CC [M]  /home/hoover/src/hid-tmff2/src/tmt500rs/hid-tmt500rs.o

/home/hoover/src/hid-tmff2/src/tmt500rs/hid-tmt500rs.c: In function ‘t500rs_send_condition_packet’:
/home/hoover/src/hid-tmff2/src/tmt500rs/hid-tmt500rs.c:481:16: error: implicit declaration of function ‘t500rs_send_hid’; did you mean ‘t300rs_send_int’? [-Werror=implicit-function-declaration]
  481 |         return t500rs_send_hid(t500rs, buf,
      |                ^~~~~~~~~~~~~~~
      |                t300rs_send_int

```

and several others following this error which look like version mismatches in function names: 

```
error: implicit declaration of function ‘t500rs_build_r03_constant’; did you mean ‘t500rs_build_r05_condition’? [-Werror=implicit-function-declaration]
  516 |         t500rs_build_r03_constant(r3, code, scaled_level);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~

```

Please forgive my ignorance as this is probably the wrong channel to report these issues in... 

All the best, 

Uwe

hoover67:
Good morning @Kimplul & thanks for the suggestions. 

I got a bit further today by installing Assetto Corsa Rally (a rather new sim) running on "Proton experimental" and I'm happy to report that FFB works nicely and more or less OOTB with this title, so it's definitely some problem with rfactor2. 

I also have eurotruck simulator 2, would it be helpful to see if this title works on my end? 

I tried using these launch options in steam with rfactor2: 

PROTON_ENABLE_HIDRAW=1 SDL_JOYSTICK_HIDAPI=0 

and also went for older Proton versions like 9.0, sadly all to no avail. The wheel & pedals work fine in rfactor2, but not ffb. 

I also tried disabling "SteamInput" which did not make a difference, either. 

During one test session I noticed some "gain set successfully" from the ffb kernel driver (quite high at 55700 or thereabouts), but no force registered on the wheel. So it looks that rf2 is trying to talk to the wheel somehow, just not correctly maybe?

```
Feb 21 10:25:38 dirac kernel: tmff2 0003:044F:B65E.000E: FFB: set_gain 55703 -> device 216
Feb 21 10:25:38 dirac kernel: tmff2 0003:044F:B65E.000E: FFB: Gain set successfully
Feb 21 10:25:51 dirac kernel: tmff2 0003:044F:B65E.000E: FFB: set_gain 55703 -> device 216
Feb 21 10:25:51 dirac kernel: tmff2 0003:044F:B65E.000E: FFB: Gain set successfully
Feb 21 10:27:13 dirac kernel: tmff2 0003:044F:B65E.000E: FFB: set_gain 55703 -> device 216
Feb 21 10:27:13 dirac kernel: tmff2 0003:044F:B65E.000E: FFB: Gain set successfully
Feb 21 10:32:08 dirac kernel: tmff2 0003:044F:B65E.000E: FFB: set_gain 55703 -> device 216


```
Another weird thing is that "oversteer" claims I don't have permissions to change the wheel setup when running it as a normal user; it works fine (obviously) using "sudo", but I'm not sure how the application is supposed to work. The permissions on the /dev/input/event* devices look ok and my user is also in the "input" group.

 Should the wheel radius change immediately when setting it in oversteer? Also, adding a centering force in oversteer makes no difference to how the wheel feels when no simulation is running and just "oversteer" is active. 

I cannot save any profile, however the "tests" section works fine and takes the t500rs through its motions. I am running 0.8.2, maybe I should try cloning the latest repo state (though the last change seems to have been in 2024). 

Again sorry for the noise in this thread, I'll try to register a proper "issue" once it's possible to compile the latest driver state from the regular repo again. 

If there's anything I can provide to help with debugging this feel free to let me know! 

All the best, 

Uwe

cazzoo:
After numerous tests, I haven't found a solution for that issue that happens for both rfactor2 and RRRE. I believe that may be related to Wine/Proton and could be managed by a fix introduced here (but don't know what).
The python script you shared is just resetting the wheel, letting it re-enumerating, and doing mode switch (from boot mode to normal mode). This is handled by the hid-tminit driver, loaded before the hid-tmff driver (which loads then T500RS part). With the latest code, I believe that if you just unplug/replug the wheel, the behavior should be the same as the python script ran.

@Kimplul have you ever heard about that issue. When entering in-game (3D) first, FFB works, when getting to menu FFB drops, and then when getting back to in-game it doesn't work anymore.
The only solution to get FFB back is to re-enumerate the wheel for the game to start effects.

I noticed already during implementation that the games usually don't send start commands. So the wheel is playing once after enumeration, then the packets are being updated in-place (which work for most of the games, but RRRE and Rfactor2).
If the game would eventually stop the effect, there would be then no play effect triggered back, leading to loss of forces.

All of this is assumptions, but this is what I roughly feel

cazzoo:
I nailed the issue a bit down and found some more clues. So confirming the issue happens for both games, I found that the game (when doing alt+tab or get back to pits) is stopping effects. When entering back in race, somehow, the effects are not started back..
Here are some debug logs I captures while this issue happens:
```
## pressed alt+tab
[32861.719208] tmff2 0003:044F:B65E.0020: FFB: Effect 0 STOP requested (type=81)
[32861.719318] tmff2 0003:044F:B65E.0020: FFB: Effect 1 STOP requested (type=81)
[32861.719417] tmff2 0003:044F:B65E.0020: FFB: Effect 2 STOP requested (type=85)
[32861.719507] tmff2 0003:044F:B65E.0020: FFB: Effect 3 STOP requested (type=83)
[32861.719592] tmff2 0003:044F:B65E.0020: FFB: Effect 4 STOP requested (type=82)
[32861.719980] tmff2 0003:044F:B65E.0020: FFB: Effect 0 STOP requested (type=81)
[32861.720071] tmff2 0003:044F:B65E.0020: FFB: Effect 0 STOP requested (type=81)
[32861.720075] tmff2 0003:044F:B65E.0020: FFB: Effect 1 STOP requested (type=81)
[32861.720078] tmff2 0003:044F:B65E.0020: FFB: Effect 2 STOP requested (type=85)
[32861.720081] tmff2 0003:044F:B65E.0020: FFB: Effect 3 STOP requested (type=83)
[32861.720084] tmff2 0003:044F:B65E.0020: FFB: Effect 4 STOP requested (type=82)
[32861.720088] tmff2 0003:044F:B65E.0020: FFB: Effect 0 STOP requested (type=81)
[32861.720091] tmff2 0003:044F:B65E.0020: FFB: Effect 1 STOP requested (type=81)
[32861.720094] tmff2 0003:044F:B65E.0020: FFB: Effect 2 STOP requested (type=85)
[32861.720097] tmff2 0003:044F:B65E.0020: FFB: Effect 3 STOP requested (type=83)
[32861.720099] tmff2 0003:044F:B65E.0020: FFB: Effect 4 STOP requested (type=82)
[32861.720310] tmff2 0003:044F:B65E.0020: FFB: Effect 0 STOPPED (type=81)
[32861.723410] tmff2 0003:044F:B65E.0020: FFB: Effect 1 STOPPED (type=81)
## got back ingame (alt+tab), no FFB
[32861.723698] tmff2 0003:044F:B65E.0020: FFB: Effect 0 START requested (count=2147483647, type=81)
[32861.723930] tmff2 0003:044F:B65E.0020: FFB: Effect 1 START requested (count=2147483647, type=81)
[32861.724150] tmff2 0003:044F:B65E.0020: FFB: Effect 2 START requested (count=2147483647, type=85)
[32861.724427] tmff2 0003:044F:B65E.0020: FFB: Effect 3 START requested (count=2147483647, type=83)
[32861.739411] tmff2 0003:044F:B65E.0020: FFB: Effect 2 STARTED (type=85, count=2147483647)
[32861.755421] tmff2 0003:044F:B65E.0020: FFB: Effect 3 STARTED (type=83, count=2147483647)
[32861.771467] tmff2 0003:044F:B65E.0020: FFB: Effect 4 STOPPED (type=82)
```

So I tested adding an extra re-init script (manually triggered) : Stop work handler, Reset all effect states and then Reinitialize wheel (tmff->open), and that worked!

So I don't exactly know if there's a bug to fix in Wine/Proton, but that workaround worked. I'm trying to get it added to the base driver but I'm not entirely convinced this is the right place, especially knowing this is shared code with other wheel bases...

Kimplul:
>  So I tested adding an extra re-init script (manually triggered) : Stop work handler, Reset all effect states and then Reinitialize wheel (tmff->open), and that worked!

> So I don't exactly know if there's a bug to fix in Wine/Proton, but that workaround worked. I'm trying to get it added to the base driver but I'm not entirely convinced this is the right place, especially knowing this is shared code with other wheel bases...

I don't think that's a good approach. From the logs it looks like effects 0-3 are all requesting a start, which the driver should be able to handle just fine. Not sure why the game seems to be sending multiple stop signals, but that shouldn't matter.

Looks like all the messages are printed within ~50 milliseconds. Is it maybe possible that the driver handles the STOP requests, and then ignores the START requests because they arrived 'too soon'?

Kimplul:
> When a game sends START requests while the work handler is processing effects, a race condition can occur where the START flag is cleared in the first loop but the effect is not actually started until the second loop, causing the effect to never start.
The solution is to batch process the events in a 3 steps sequence.

That doesn't make sense to me. Setting and clearing the flags is done within a spinlock, where exactly would the race condition be?

If you mean that the worker sees a STOP, and while handling the STOP, a START arrives, I could see that the START is not picked up. However, since the worker is running, `delayed_work_pending()` should return false, and the work handler should run a second time, triggering the START. Maybe check if the work handler stops running after an alt-tab? That would be my preferred solution, you solution seems to imply that both REQUEST_START and REQUEST_STOP would be set, which should never be the case.

How is the interrupt/hid messaging stuff related? Is the interupt just slower, meaning a greater chance of mismatched start/stop?

cazzoo:
@hoover67 you have to do git fetch cazzoo origin (commands are not correct), then checkout branch "hardening/fix-multi-stop-start-race-condition" from cazzoo origin as well, and then build&install driver.
Recommended, every time you checkout a branch, to do git pull

hoover67:
@cazzoo  Looking great! Just tried rfactor2 with the latest version of your driver including the race condition workaround and I was able to enter the track several times without the wheel losing FFB effects... thanks so much guys for all your hard work.

hoover67:
Hi guys, 

just a quick note to let you know that FFB also works with the RallySimFans Richard Burns Rally mod. 

Also, after updating to Linux Mint 22.3 (was running .0 before), I no longer have to run the tmdrv script above after plugging in the wheel in order for oversteer to detect it correctly (I still have to launch oversteer as root though). 

All the best, 

Uwe

hoover67:
Hi folks, 

I just tried the june-hardening branch and lost all ffb effects on my wheel (in rfactor2, haven't tried other sims yet) except for a rather strong centering force. 

I installed the new version as per instructions in the readme after a full "clone" of the repo. 

i then reverted back to the old driver (thankfully I had a backup) and everything worked as before. 

All the best, 

Uwe

thefahre:
Hi all, 

I just tested the driver with /hardening/fix-multi-stop-start-race-condition branch (as @hoover67 said the /hardening/June-pass to not work for him) and force feedback seems to be working, however the force feedback itself feels a bit off, I don't know how to descibe it other than it has very exaggerated force near center yet almost nothing but centering spring when turning. Tried changing settings via oversteer but also doesn't seem to make much of a difference. The game I use to test is Assetto Corsa and Assetto Corsa Competizione.

I'm also unable to validate pedal functionality as I am using Thrustmaster T-LCM pedal directly via USB.

I will try the /hardening/June-pass branch next just to feel the difference.

cazzoo:
Hello both,

Thank you for your interest, I am currently working to update the driver and make it work more alike the T300RS and other variants, having effects that lasts length with driver terminating effect.
I will let you know.
Also, if you have double boot, and could capture real session, I'd be really pleased to get them. Thank you

hoover67:
Hey cazzoo, thanks for the info. I still have a win10 partition on my gaming rig somewhere, I have no idea if it still boots up though. 

What exactly would you be interested in when you're asking us to capture a session?

Cheers, Uwe

cazzoo:
@hoover67 , I am looking at any game session captures, so I can verify protocol is working as implemented.

Regarding new changes, if you are interested into the latest changes, you can checkout branch [session/agent_a361d90b-d010-4399-8b16-7988ad63f2f8](https://github.com/cazzoo/hid-tmff2/tree/session/agent_a361d90b-d010-4399-8b16-7988ad63f2f8) pulling at it, building and replacing the system driver. Happy to get your feedback on it

thefahre:
hi @cazzoo, would a Wireshark capture in bare metal Windows via USBPcap be sufficient? I'm planning to capture a practice session in Assetto Corsa, preparing for a club league endurance event.

thefahre:
Hi @cazzoo, following is capture from my T500 RS playing Assetto Corsa in Windows: [Google Drive link](https://drive.google.com/file/d/1YmYorQjLbkDT_YRjZ971x_g70Z-pNaf1/view?usp=sharing)

Also attached are settings of the wheel and the game.
Screenshot 2026-07-18 120853
Screenshot 2026-07-18 120819
Screenshot 2026-07-18 120755
Screenshot 2026-07-18 120745

hoover67:
I checked this morning and my win10 install is no longer operational, so I won't be able to provide a USB capture of an rf2 session on short notice... sorry. 

All the best, Uwe

hoover67:
Hi all, 

I managed to revive my win10 partition from the dead and did a little rfactor2 session at Kyalami 1976 in the Brabham BT24 by ChiefWiggum on our practise server (all content is freely available from the steam workshop). 

The pcap file can be found here: 

https://www.icloud.com/iclouddrive/075xh2eBnyMsPO4P34eEm_lmw#t500rs%5Ff1%5Fwheel%5Frfactor2%5Ff1%5F1967%5Fbt24%5Fkyalami%5F1976%5Fonline

I did not do any specific filtering as the t500 wheel is the only device connected to this physical port. (front-facing USB3 port). I had to disable steam input for rfactor2 as the title would not detect any input from my wheel. I left all values (ffb, multipliers etc) at default. 

The session includes plugging in the wheel right at the beginning of the capture (initial rotation of the hardware wheel over the entire range by the firmware). 

Please let me know if any other info is required from my end @cazzoo 

All the best & thanks again for your work on this project! 

Uwe

KORraNpl:
hey guys, in case help is needed, I'll do my best. So far I've been dual-booting to play racing games and just tried to use [session/agent_a361d90b-d010-4399-8b16-7988ad63f2f8](https://github.com/cazzoo/hid-tmff2/tree/session/agent_a361d90b-d010-4399-8b16-7988ad63f2f8) branch (I see it's being actively worked on :+1: ), but while the wheel is detected, it jumps by a lot near the center, so it's undrivable in ACC. I've downloaded the most recent Overdrive app and there my T500 RS is not detected at all.

I'm using Kubuntu 24.04 (7.0.0-30-generic), installed module manually via `make && sudo make install`.

kakra:
> but while the wheel is detected, it jumps by a lot near the center

Did you try the latest git version and removed old udev rules that may be pre-existing? The current git version has much higher wheel precision. Maybe said branch should be rebased?

thefahre:
Gave my feedback to latest changes in #218, commenting here to serve as a link.

hoover67:
Quick update: Tested the new driver in RSF RBR today, appears to work nicely there, too.

cazzoo:
Closing this PR, the other one (https://github.com/Kimplul/hid-tmff2/pull/218) does have fresher codebase and better support.

Issue #198, opened 2026-03-22
Thrustmaster T248R Support


renfordt:
I'm currently struggeling to set up my T248R.

OS: CachyOS

I have installed hid-tmff2 and also oversteer via AUR both seem to work in general but instead of a T248R a T300RS is selected in the compatibility mode, with no other selections in that dropdown possible. I have the T3PM pedals.

The steering wheel seems to work in general, but the buttons don't work at all. 

Clutch, brakes and accelerator  are swapped in oversteer and also don't react properly (only 0, 0,5 or 1; not really floating)
Accelerator is clutch, breaks are accelerator and clutch is break.

If you give me some hints how to get the correct values, I can also create a pull request and make it working (hopefully).

Kimplul:
Hello, thanks for the report. I hadn't heard of the T248R before, I just have the 'regular' T248 so this is potentially new territory. Could you post the output of `sudo dmesg -w` when connecting the wheel?

renfordt:
Of course here it is:

```
[19473.029585] usb 1-2.4: new full-speed USB device number 21 using xhci_hcd
[19473.257847] usb 1-2.4: New USB device found, idVendor=044f, idProduct=b66d, bcdDevice= 2.49
[19473.257855] usb 1-2.4: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[19473.257857] usb 1-2.4: Product: Thrustmaster Racing Wheel FFB
[19473.257859] usb 1-2.4: Manufacturer: Thrustmaster
[19473.305958] input: Thrustmaster Thrustmaster Racing Wheel FFB as /devices/pci0000:00/0000:00:02.1/0000:03:00.0/0000:04:0c.0/0000:0b:00.0/usb1/1-2/1-2.4/1-2.4:1.0/0003:044F:B66D.0018/input/input40
[19473.306070] tmff2 0003:044F:B66D.0018: input,hiddev96,hidraw1: USB HID v1.11 Gamepad [Thrustmaster Thrustmaster Racing Wheel FFB] on usb-0000:0b:00.0-2.4/input0
[19473.311810] tmff2 0003:044F:B66D.0018: firmware version 1 might be too old, consider updating
[19473.311811] tmff2 0003:044F:B66D.0018: note: this has to be done through Windows.
[19473.317811] tmff2 0003:044F:B66D.0018: force feedback for T300RS
```

Kimplul:
> [...]044F:B66D[...]

[...]

PR #205, opened 2026-05-28
T598 kimplul

dapperflapper:
This patch adds initial force feedback support for the Thrustmaster T598 
racing wheel.

The T598 shares a USB product ID (TMT248_PC_ID) with the T248, so device 
detection is done by reading bcdDevice from the USB descriptor — 0x0598 
identifies the T598, falling back to the T248 path otherwise.

Changes:
- Added bcdDevice-based detection in tmff2_probe() to distinguish T598 from T248
- Added TMT598_PC_ID to the hid_device_id table
- Added new driver file src/tmt598/hid-tmt598.c implementing the T598 API:
  - Correct send_buffer length (63 bytes) and indexing
  - Correct ff_field assignment from report->field
  - Range clamped to 140–1080 degrees
  - Full force feedback effect support (constant, ramp, spring, damper, 
    friction, inertia, periodic, sine, triangle, square, saw up/down, 
    autocenter, gain)
  - Reuses t300rs play/upload/update/stop effect implementations

Tested on a physical Thrustmaster T598 on Linux.


Kimplul:
Hi, thanks for the PR. `bcdDevice` is a nice catch for the T598, seems it could be used for detecting other wheels. The T248 ID / 'Advanced Racer' is used for quite a few models, there's been some work gathering info about the Thrustmaster API itself to see if it could be used to ask the wheel itself what it thinks it is, see https://github.com/Kimplul/hid-tmff2/issues/198#issuecomment-4364512716, this might help with that.

There's a rather strong smell of AI here, did you use any? It's generally a polite thing to mention, and might change what approach to reviewing I take.

Issue #122, opened 2024-10-16
Small "deadzone" when changing wheel direction (T300)


kiiro94:
First of all, thanks for the driver, it works amazingly.

One little thing that I noticed, is that whereas on Windows the wheel is responsive even with micro movements (down to the 0.1% on the control panel), on Linux there's a small "deadzone" if and only if I change the direction from left to right , or right to left.
As soon as I change the direction of travel of the wheel, there's a small deadzone.

I can kinda show it to you using Assetto Corsa content manager: I'm moving the wheel in the most linear way possible, and you can see a small pause in the input movement when the wheel goes from left to right, or right to left.

It feels like a couple degrees max of deadzone, but you can feel it in games, especially when going fast on a straight.

https://github.com/user-attachments/assets/5223efe9-a85e-4134-86c9-e9acb7de3a0c

kiiro94:
Fixed it by running 
evdev-joystick --e /dev/input/by-id/usb-Thrustmaster_Thrustmaster_Racing_Wheel_FFB-event-joystick --d 0 --f 0

It works until next reboot/unplug of the wheel, and now the wheel behaves exactly like on Windows.

kiiro94:
> Glad you got it working. Just to check, is this with the `udev` rule installed as outlined in the README?
> 
> https://github.com/Kimplul/hid-tmff2?tab=readme-ov-file#common-issues-and-notes

The udev rule did not work on my system (Fedora 41 beta). Dunno if I did something wrong. I had to execute that particular command to fix the deadzone.

kiiro94:
> I updated the `udev` part of the README, would you mind checking if it works for you?

No, I copied all the rules just to be sure, but they are not working.

Kimplul:
Interesting. Which wheel are you using? Does `dmesg` show the product ID when you plug the wheel in?

Issue #42, opened 2022-06-22
large dead zone in the center


sayapinsergej:
I got a very large dead zone in the center. I created a file /etc/udev/rules.d/99-joydev.rules. What value should be changed to --deadzone 0?

Kimplul:
`--deadzone 0` should apply to all axes, if there's still a deadzone it's likely something in game. To answer your question though, the `X` axis is the wheel.

sayapinsergej:
In the oversteer program, the steering wheel rotation is displayed correctly. I have jstest-gtk installed, and it shows a large dead zone in the center

Kimplul:
Yep, Oversteer uses the `evdev` subsystem, which is the subsystem whose deadzone is set to zero by the udev rule. `jstest-gtk` uses the older `joydev` subsystem, which is at this point obsolete and I'm not aware of any games that use it.

You can read up on the older interface and how to handle its deadzones here:
https://wiki.archlinux.org/title/Gamepad#Joystick_API_deadzones

Rudicito:
Hi,
I have the same problem, a large dead zone in the center and the correct rotation in the oversteer program, but the thing is that is happen in wine with lutris even with thrustmaster driver install and thethe dead zone set to 0 with the registry key. I can see it in the wine control panel and it's the same thing in the game. It's the same thing on steam. I don't understand how other persons can play games with a thrustmaster wheel with wine. I'm on Nobara 36. I don't think that Wine and steam using joydev subsystem. Can somebody help me. Thanks in advance.

Kimplul:
Wine uses SDL2, which uses the `evdev` subsystem, you're correct.

Could you specify which wheel you're using and in which mode it is, and which games have this dead zone? All Steam games?

Rudicito:
Hi, Thank you for responding.
I'm using the t300rs gt edition in ps3 normal mode. The only game that works for me is RaceRoom Racing Experience on steam. I managed to go into the control panel of the steam proton wine of the game and I see that the wheel rotation is correct. My question is where I have to look to find a config file or something like that, to be able to do the same in the wine-ge-custom in lutris [https://github.com/GloriousEggroll/wine-ge-custom](url). The game I mainly want to play in lutris with my wheel is Richard Burns Rally. Thanks in advice.

Kimplul:
> I managed to go into the control panel of the steam proton wine of the game and I see that the wheel rotation is correct.

[...]

Issue #34, opened 2022-01-19
libbeamng.| Steering wheel drivers didn't provide any FFB resolution information. Defaulting to 65536 steps and wheel getting "stuck" sometimes in beamng


kevenwyld:
First off, thank you very much for your hard work on this project.  I'd love to buy you a beer or something if you accept donations for your time.  We pay so much for these devices and games but the parts that really stitch together Linux gaming are often small projects like this and the people behind them.

I've just purchased a used T300 and have been testing this out with several games.  Everything is mostly working great however in beamng, which is the game I spend the most time in, I've experienced two issues:

1. The game reports `libbeamng.| Steering wheel drivers didn't provide any FFB resolution information. Defaulting to 65536 steps`.  This doesn't seem to be so much of an issue, FFB is functional, but I'm curious what exactly this might mean and thought it might be something you should be aware of.
2. When playing beamng I've noticed that occasionally the wheel will violently move to a 90 degree position and get "stuck".  Usually after a large FFB event like a jump.  If I turn very hard back to center it will "unstick" and seems to be OK for a while until the next time.  This doesn't happen often, between 0 and 2 times every time I play the game.

I know that beamng is running through proton and there are a lot of variables there with all the compatibility layers but it would be interesting to hear if you've encountered this and if you have any ideas.  Also I thought these might be related so I put them in the same issue.  Let me know if I should open separate ones.

Thanks again!

Kimplul:
Thanks for the report. I have no idea what FFB resolution the game is talking about, I'll have to see if I can find some info on it. About getting stuck, I had to turn on inverted FFB from the game's FFB settings for it to work properly. Not sure if something could be changed to make the game figure out the direction by default, but at least it's an easy fix.

kevenwyld:
I feel like I'm just not understanding how the damping is suppose to work maybe, this is my first high output wheel (had a G27 before) so I'm sorry if it turns out that I'm just not using it correctly.

After more experimentation I'm able to reproduce this pretty much immediately by taking a sharp turn and then quickly bringing the wheel back to center and stopping the vehicle.  Here's some videos of the issue.  I disabled the in game FFB setting `reduce strength at park` to make the issue appear more easily in the second video, but it's the same thing that happens occasionally in game.  In the first video the in game FFB settings are default except for `inverted` which is checked.

Files are too large for github so I put them on vimeo, sorry about the one-handed vertical video, it's the best I could do trying to hold everything:

https://vimeo.com/667966036

[...]

Issue #151, opened 2025-03-17
T-GT II - Enchantments


samsunix:
Hello again,

I would like to request little T-GT2 enchantment.
Following controls highlighted on [picture](https://imgur.com/1UWF2Zg) are not functional and not detected by oversteer.

They are not absolutely crucial but 'nice to have'. Let me know what you need and I will do my best to provide.

Kimplul:
Presumably the T-GT II has an `rdesc` that's different from the one(s) in https://github.com/Kimplul/hid-tmff2/blob/a9312ead5720922e1c06b541ea48b6db3f289d36/src/tmt300rs/hid-tmt300rs.c#L87, so first order of business would probably be to try and get hold of it. With the driver uninstalled (or disabled or whatever, as long as it doesn't get loaded) you should be able to query the raw `rdesc` from the wheel with
```
cat /sys/kernel/debug/hid/XXXX\:VID\:PID.YYYY/rdesc
```

I believe `XXXX = 044f` and `YYYY = b66e` but you might have to dig around a bit (there are other ways to get the `rdesc`, for example capturing packets with WireShark but I think the above method is probably the easiest).

With a 'proper' `rdesc`, we would then need to figure out when we're talking to a T-GT II versus a regular T300. I would expect that information to be available via `t300rs_get_attachment()`, you might want to temporarily add in some print statements [here](https://github.com/Kimplul/hid-tmff2/blob/a9312ead5720922e1c06b541ea48b6db3f289d36/src/tmt300rs/hid-tmt300rs.c#L1409) (again, there are other ways to get the information like sending raw USB packets from the command line but I think the 'hack' above is probably easiest).

samsunix:
I dunno how to do pull request but maybe better if you check these entries first before I try and compile driver again because I have no clue what the heck I am doing.
```
0x09, 0x40,  /* Usage (Knob 1 Rotation) */
0x15, 0x00,  /* Logical minimum (0) */
0x26, 0xff, 0x03,  /* Logical maximum (1023) */
0x75, 0x10,  /* Report size (16) */
0x95, 0x01,  /* Report count (1) */
0x81, 0x02,  /* Input (Variable, Absolute) */

0x09, 0x41,  /* Usage (Knob 1 Button Press) */
0x15, 0x00,  /* Logical minimum (0) */
0x25, 0x01,  /* Logical maximum (1) */
0x75, 0x01,  /* Report size (1) */
[...]

PR #211, opened 2026-08-08
fix force feedback arithmetic and wheel input filtering

kakra:
## Summary

This series contains four independent correctness changes in force feedback and wheel input handling:

- restrict the configured driver gain to its documented unsigned 16-bit range
- avoid signed integer overflow when combining application and driver gain
- preserve the full signed effect range during direction scaling
- remove generic HID fuzz and flat filtering from wheel and pedal axes

The commits are intentionally kept separate so each change remains bisectable and can be reviewed or reverted independently.

No gain or rotation-range defaults are changed.

## Gain configuration and scaling

The configured driver gain is documented as an unsigned 16-bit value, but was stored and parsed as `int`. Values outside `0..65535` were accepted and silently clamped later.

The configured gain is now stored as `u16`. Module-parameter and sysfs values outside the documented range are rejected as invalid user input.

Both `FF_GAIN` and the driver gain use the full unsigned 16-bit range. Their product can exceed `INT_MAX`, so the existing signed multiplication invokes undefined behaviour for common configurations such as 75 percent application gain combined with 75 percent driver gain.

Gain multiplication now uses a `u32` intermediate, which can hold the product of two full-range 16-bit values. The initial and sysfs paths pass the already validated gain directly instead of using redundant multiply-divide expressions which could overflow before cancelling out.

## Signed effect scaling

The driver uses a symmetric fixed-point sine range of `-32767..32767`. Scaling the valid input value `-32768` by `-32767` therefore produces `32768`.

The old code stored that result in `s16` before completing the effect conversion, wrapping it to `-32768`. At full scale this could reverse the force direction instead of producing the expected positive force.

Direction-scaled values are now kept in `s32` until the destination range is known. Periodic magnitudes are made positive in the wider type and clamped to `S16_MAX`, preventing both the endpoint wrap and the subsequent invalid headroom calculation.

Constant and ramp effects retain their wider intermediates until their final conversion as well.

## Input filtering

The generic HID input setup derives fuzz and flat values from the logical axis range. On the tested T300 this produces:

```text
ABS_X: fuzz=255 flat=4095
pedals: fuzz=3 flat=63
```

The steering values suppress and smooth small corrections before an application receives them. This is particularly noticeable when changing steering direction.

flat represents a centered deadzone, which is also unsuitable for pedals whose rest position is an endpoint. Pedal fuzz can suppress small changes near an endpoint, causing the reported value to remain slightly short of the hardware endpoint.

The input setup now sets fuzz=0 and flat=0 for every available wheel and pedal axis on devices handled by hid-tmff2. Missing axes are skipped through the input device's absbit mask.

This follows the existing hid-universal-pidff precedent of overriding generic HID filtering for high-resolution racing devices.

## Related issues

- Addresses the underlying driver behaviour reported in #122. The reporter confirmed that setting both fuzz and flat to zero eliminates the small deadzone when changing steering direction.
- Likely addresses the force-direction wrap described in #34. That issue reports that reaching the positive force limit can suddenly jerk the wheel in the opposite direction, and its game-side workaround clamps the force just below that endpoint. This matches the signed 32768 to -32768 wrap fixed here, but still needs runtime confirmation.
- May address the kernel-generated portion of the deadzone reports in #42. That issue also contains joydev-, Wine- and runtime-specific cases, so this series does not claim to resolve every symptom discussed there.

## Testing

Completed on a T300RS GT:

- make and make W=1 against Linux 6.18.38
- git diff --check
- numerical validation across the complete signed effect input range
- loaded and tested the module built from this exact branch
- verified fuzz=0 and flat=0 on the steering and pedal axes
- regression-tested steering, pedals and Force Feedback in ETS2 under Proton
- exercised Constant, Ramp, Periodic, Spring, Damper and Friction effects
- exercised 100 percent driver master gain during normal gameplay

Damper and Friction effects were accepted without an observed error, although their individual contribution was subjectively subtle. Damper appeared to reduce wheel run-on, consistent with velocity-dependent damping.

Still pending before marking this PR ready:

- regression-test Forza Horizon 5 with the module built from this exact branch

Deferred hardware validation:

- related Thrustmaster wheel models are not locally available and will require testing by the maintainer or community
- additional master-gain values may be tested, but the scaling arithmetic has already been validated numerically over its full input range

### DIFF
--- a/src/hid-tmff2.c
+++ b/src/hid-tmff2.c
@@ -48,6 +48,11 @@ module_param(gain, ushort, 0);
 MODULE_PARM_DESC(gain,
 		"Level of gain (0-65535)");
 
+static u16 tmff2_scale_gain(u16 value)
+{
+	return (u32)value * gain / GAIN_MAX;
+}
+
 static spinlock_t lock;
 
 static struct tmff2_device_entry *tmff2_from_hdev(struct hid_device *hdev)
@@ -244,7 +249,7 @@ static ssize_t gain_store(struct device *dev,
 
 	gain = value;
 	if (tmff2->set_gain) /* if we can, update gain immediately */
-		tmff2->set_gain(tmff2->data, (GAIN_MAX * gain) / GAIN_MAX);
+		tmff2->set_gain(tmff2->data, gain);
 
 	return count;
 }
@@ -268,7 +273,7 @@ static void tmff2_set_gain(struct input_dev *dev, uint16_t value)
 		return;
 	}
 
-	if (tmff2->set_gain(tmff2->data, (value * gain) / GAIN_MAX))
+	if (tmff2->set_gain(tmff2->data, tmff2_scale_gain(value)))
 		hid_warn(tmff2->hdev, "unable to set gain\n");
 }
 
@@ -644,7 +649,7 @@ static int tmff2_wheel_init(struct tmff2_device_entry *tmff2)
 	/* set defaults wherever possible */
 	if (tmff2->set_gain) {
 		ff->set_gain = tmff2_set_gain;
-		tmff2->set_gain(tmff2->data, (GAIN_MAX * gain) / GAIN_MAX);
+		tmff2->set_gain(tmff2->data, gain);
 	}
 
 	if (tmff2->set_autocenter)
--- a/src/hid-tmff2.c
+++ b/src/hid-tmff2.c
@@ -807,6 +807,26 @@ static void tmff2_remove(struct hid_device *hdev)
 	kfree(tmff2);
 }
 
+static int tmff2_input_configured(struct hid_device *hdev,
+				  struct hid_input *hidinput)
+{
+	struct input_dev *input = hidinput->input;
+	int axis;
+
+	if (!input->absinfo)
+		return 0;
+
+	for (axis = ABS_X; axis <= ABS_BRAKE; axis++) {
+		if (!test_bit(axis, input->absbit))
+			continue;
+
+		input_abs_set_fuzz(input, axis, 0);
+		input_abs_set_flat(input, axis, 0);
+	}
+
+	return 0;
+}
+
 static const struct hid_device_id tmff2_devices[] = {
 	/* t300rs and variations */
 	{HID_USB_DEVICE(USB_VENDOR_ID_THRUSTMASTER, TMT300RS_PS3_NORM_ID)},
@@ -831,6 +851,7 @@ static struct hid_driver tmff2_driver = {
 	.probe = tmff2_probe,
 	.remove = tmff2_remove,
 	.report_fixup = tmff2_report_fixup,
+	.input_configured = tmff2_input_configured,
 };
 module_hid_driver(tmff2_driver);
 
--- a/src/hid-tmff2.c
+++ b/src/hid-tmff2.c
@@ -43,8 +43,8 @@ MODULE_PARM_DESC(alt_mode,
 		"Alternate mode, eg. F1 mode");
 
 #define GAIN_MAX 65535
-int gain = 40000;
-module_param(gain, int, 0);
+u16 gain = 40000;
+module_param(gain, ushort, 0);
 MODULE_PARM_DESC(gain,
 		"Level of gain (0-65535)");
 
@@ -230,14 +230,15 @@ static ssize_t gain_store(struct device *dev,
 		struct device_attribute *attr, const char *buf, size_t count)
 {
 	struct tmff2_device_entry *tmff2 = tmff2_from_hdev(to_hid_device(dev));
-	unsigned int value;
+	u16 value;
 	int ret;
 
 	if (!tmff2)
 		return -ENODEV;
 
-	if ((ret = kstrtouint(buf, 0, &value))) {
-		dev_err(dev, "kstrtouint failed at gain_store: %i", ret);
+	ret = kstrtou16(buf, 0, &value);
+	if (ret) {
+		dev_err(dev, "failed to parse gain: %d\n", ret);
 		return ret;
 	}
 
@@ -251,7 +252,7 @@ static ssize_t gain_store(struct device *dev,
 static ssize_t gain_show(struct device *dev,
 		struct device_attribute *attr, char *buf)
 {
-	return scnprintf(buf, PAGE_SIZE, "%i\n", gain);
+	return scnprintf(buf, PAGE_SIZE, "%u\n", gain);
 }
 static DEVICE_ATTR_RW(gain);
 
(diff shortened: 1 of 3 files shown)


src/hid-tmff2.c
```diff
@@ -812,6 +812,50 @@ static void tmff2_remove(struct hid_device *hdev)
...
+	case TMT300RS_PS4_NORM_ID:
+		break;
+	default:
+		return 0;
+	}
+
+	if (!input->absinfo)
+		return 0;
+
+	for (i = 0; i < ARRAY_SIZE(t300_axes); i++) {
```
Kimplul: Comment references `hid-universal-pidff`, which just loops over the absolute axes, while this creates a duplicate list of axes. Seems like the `hid-universal-pidff` way is a bit cleaner, any reason why this was done the way it was?

https://elixir.bootlin.com/linux/v6.18.40/source/drivers/hid/hid-universal-pidff.c#L127

src/hid-tmff2.c
```diff
@@ -812,6 +812,50 @@ static void tmff2_remove(struct hid_device *hdev)
...
+		ABS_THROTTLE,
+		ABS_RUDDER,
+		ABS_WHEEL,
+		ABS_GAS,
+		ABS_BRAKE,
+	};
+	struct input_dev *input = hidinput->input;
+	unsigned int i;
+
+	switch (hdev->product) {
```
Kimplul: Don't think this is necessary, apparently the [T818](https://github.com/Kimplul/hid-tmff2/pull/143#discussion_r1929798039) benefits from removing fuzzing and if you see this benefitting the T300, that covers the belt and gear systems in use (that I'm aware of) so presumably all wheels benefit from this.

src/hid-tmff2.c
```diff
@@ -240,10 +245,15 @@ static ssize_t gain_store(struct device *dev,
 		dev_err(dev, "kstrtouint failed at gain_store: %i", ret);
 		return ret;
 	}
+	if (value > GAIN_MAX) {
```
Kimplul: Nitpick/discussion thing:

I see the point of the change, the gain handling is currently a bit off, but I wonder if it might make more sense to set the `gain` parameter to a `u16` and use `kstrtou16`? That would encode the intended usage a bit better and an invalid gain parameter becomes user error.

Kimplul:
Thanks for the PR, appreciate it! I commented on a couple minor things, although really only https://github.com/Kimplul/hid-tmff2/pull/211#discussion_r3740744811 is a functional change, the other ones are kind of personal preference I guess.

kakra:
Thanks for the feedback. I addressed your commets with fixup commits for now. Before merge, I'll rebase and reorder the commits.

src/hid-tmff2.c
kakra: Changed the sysfs parameter type as you suggested. It's actually the better idea, mea culpa: I didn't see it although I'm taking care of that in xpadneo a lot.

src/hid-tmff2.c
kakra: I generalized the changes to all supported models now. Additionally, axis are no longer filtered. So if we have models that support analog joysticks on the wheel, it will have no fuzz or flat neither, although it should probably do so. But at least for Proton games, it's almost always better to expose such joysticks with no fuzz or flat because Windows games to their own noise filtering and deadzone calculations, as I learned from developing xpadneo. I'm soon going to experimentally remove fuzz from xpadneo to see how it feels, and it should remove artificial input latency and improve precision. So, removing fuzz/flat from joystick axes may not actually be the worst thing to do.

src/hid-tmff2.c
kakra: Oh well, okay, commented below... I was mainly commenting about why or why not axes should be filtered. ;-)

src/hid-tmff2.c
Kimplul: > So if we have models that support analog joysticks on the wheel, it will have no fuzz or flat neither, although it should probably do so.

Ah, makes sense. I'm not aware of base wheels with joysticks, but there's a T-GT II wheel [addon](https://www.thrustmaster.com/en-us/products/gt-wheel-add-on/) with two joysticks, and an open issue about them currently not being supported: https://github.com/Kimplul/hid-tmff2/issues/151

The T-GT II is otherwise supported by this driver, it just reuses the same VID/PID as the T300 and this driver doesn't currently have a way to differentiate the two, so the joysticks are not present in the `rdesc`. Interestingly, [it seems that Thrustmaster differentiates wheels](https://github.com/Kimplul/hid-tmff2/issues/198#issuecomment-4565789561) (please ignore me being braindead later on in the thread) with the `bcdDevice` field, which is fairly straightforward to do in this driver but I'm not sure how userspace should react to it. At least SDL and Oversteer just have a list of VID/PID pairs that they use to check what kind of device they're dealing with.

kakra:
@Kimplul BTW: I have a few Proton commits pending (not related to your code, just informational): Games in Proton do not detect the wheel when cold plugged while they use the raw Gaming Input interfaces. There's a missing index reset in the interface. Also, for games using that interface, the force values seem to be sent inverted. I'm still in the process of testing those commits here before submitting them upstream to wine, so nothing to see yet. I just wanted to let you know because it looks like some issues mention exactly these bugs.

src/hid-tmff2.c
kakra: If it is not part of rdesc or somewhat mixed up, just create sub devices using `HID_QUIRK_INPUT_PER_APP` or create multiple additional synthesized devices. xpadneo has code examples of how to do that and handle it during init and teardown.

kakra:
I updated the initial PR text to reflect the current state of the code.

src/hid-tmff2.c
kakra: > Ah, makes sense. I'm not aware of base wheels with joysticks

I think Thrustmaster doesn't offer such models, only addon modules. But other vendors like Logitech and Forza may have such wheels. Thrustmaster might in the future so I filtered for axes. But in the end, Proton games usually benefit if you don't fuzz or flat joysticks, so let's just stick with no axes filtering until someone complains that joysticks don't properly center.

Kimplul:
> @Kimplul BTW: I have a few Proton commits pending (not related to your code, just informational): Games in Proton do not detect the wheel when cold plugged while they use the raw Gaming Input interfaces. There's a missing index reset in the interface. Also, for games using that interface, the force values seem to be sent inverted. I'm still in the process of testing those commits here before submitting them upstream to wine, so nothing to see yet. I just wanted to let you know because it looks like some issues mention exactly these bugs.

Cool! Please do ping me when you're sending the patches, I can relay the info to the relevant threads and see what happens.

Thanks a lot for the helpful pointers as well.

kakra:
> Please do ping me

Noted, I've set a reminder to do that when the Valve PRs are created.

Kimplul:
I checked that my T300RS and T248 behaved properly and my nitpicks were taken care of, so I went ahead and merged. Thanks!

PR #213, opened 2026-08-16
hid-tmff2: round scaled force-feedback values

kakra:
## Summary

Use the kernel's `DIV_ROUND_CLOSEST()` helper when converting and scaling force-feedback values.

This affects:

- global gain scaling;
- T300 direction and magnitude scaling;
- ConstantForce level conversion;
- periodic phase conversion;
- condition coefficients and saturation;
- condition deadband splitting;
- ramp slope and center calculation.

The calculations previously used integer division truncating towards zero. For small signed force values this can suppress effects and introduce a directional bias between positive and negative forces.

## Motivation

Low-amplitude effects are particularly important for steering wheels. In ETS2, small engine and road-surface vibrations previously felt noticeably biased towards one direction. After applying nearest-integer rounding, these effects became more detailed and symmetric during small steering corrections.

The change does not alter gain, range, device defaults or the T300 protocol layout. Intermediate calculations remain widened before division and the existing output limits are preserved.

## Testing

- Built successfully against Linux 6.18.43 in a clean temporary build tree.
- Existing `modpost` warning about the composite module's `MODULE_DESCRIPTION()` remains unchanged.
- Tested on a Thrustmaster T300RS with low-force ETS2 effects.
- Fine engine vibration and small steering corrections were subjectively more symmetric after the change.

### DIFF
--- a/src/hid-tmff2.c
+++ b/src/hid-tmff2.c
@@ -50,7 +50,7 @@ MODULE_PARM_DESC(gain,
 
 static u16 tmff2_scale_gain(u16 value)
 {
-	return (u32)value * gain / GAIN_MAX;
+	return DIV_ROUND_CLOSEST((u32)value * gain, GAIN_MAX);
 }
 
 static spinlock_t lock;
--- a/src/hid-tmff2.h
+++ b/src/hid-tmff2.h
@@ -3,6 +3,7 @@
 #define __HID_TMFF2_H
 
 #include <linux/fixp-arith.h>
+#include <linux/math.h>
 #include <linux/ktime.h>
 #include <linux/input.h>
 
--- a/src/tmt300rs/hid-tmt300rs.c
+++ b/src/tmt300rs/hid-tmt300rs.c
@@ -346,7 +346,8 @@ static uint16_t t300rs_calculate_length(uint16_t length)
 
 static s32 t300rs_scale_direction(s16 level, u16 direction)
 {
-	return (s32)level * fixp_sin16(direction * 360 / 0x10000) / 0x7fff;
+	return DIV_ROUND_CLOSEST((s32)level *
+		fixp_sin16(direction * 360 / 0x10000), 0x7fff);
 }
 
 static int16_t t300rs_calculate_constant_level(int16_t level, uint16_t direction)
@@ -354,7 +355,7 @@ static int16_t t300rs_calculate_constant_level(int16_t level, uint16_t direction
 	s32 scaled_level = t300rs_scale_direction(level, direction);
 
 	/* the Windows driver uses the range [-16385;16381] */
-	return scaled_level / 2;
+	return DIV_ROUND_CLOSEST(scaled_level, 2);
 }
 
 static void t300rs_calculate_periodic_values(struct ff_effect *effect)
@@ -375,7 +376,8 @@ static void t300rs_calculate_periodic_values(struct ff_effect *effect)
 	periodic->magnitude = min_t(s32, magnitude, S16_MAX);
 
 	/* the interval [0; 32677[ is used by the wheel for the [0; 360[ degree phase shift */
-	periodic->phase = periodic->phase * 32677 / 0x10000;
+	periodic->phase = DIV_ROUND_CLOSEST((u32)periodic->phase * 32677,
+					    0x10000);
 
 	headroom = 0x7fff - periodic->magnitude;
 	/* magnitude + offset cannot be outside the valid magnitude range, */
@@ -419,7 +421,7 @@ static int16_t t300rs_calculate_coefficient(int16_t coeff, uint16_t effect_type)
 		break;
 	}
 
-	return coeff * input_level / 100;
+	return DIV_ROUND_CLOSEST((s32)coeff * input_level, 100);
 }
 
 static uint16_t t300rs_calculate_saturation(uint16_t sat, uint16_t effect_type)
@@ -429,7 +431,7 @@ static uint16_t t300rs_calculate_saturation(uint16_t sat, uint16_t effect_type)
 	if(sat == 0)
 		return max;
 
-	return sat * max / 0xffff;
+	return DIV_ROUND_CLOSEST((u32)sat * max, 0xffff);
 }
 
 static void t300rs_calculate_deadband(int16_t *out_rband, int16_t *out_lband,
@@ -437,8 +439,10 @@ static void t300rs_calculate_deadband(int16_t *out_rband, int16_t *out_lband,
 {
 	/* max deadband value is 0x7fff in either direction */
 	/* deadband is the width of the deadzone, one direction is half of it */
-	*out_rband = clamp(offset + (deadband / 2), -0x7fff, 0x7fff);
-	*out_lband = clamp(offset - (deadband / 2), -0x7fff, 0x7fff);
+	*out_rband = clamp(offset + DIV_ROUND_CLOSEST(deadband, 2),
+				  -0x7fff, 0x7fff);
+	*out_lband = clamp(offset - DIV_ROUND_CLOSEST(deadband, 2),
+				  -0x7fff, 0x7fff);
 }
 
 static void t300rs_calculate_ramp_parameters(uint16_t *out_slope,
@@ -455,8 +459,8 @@ static void t300rs_calculate_ramp_parameters(uint16_t *out_slope,
 	end_level = t300rs_scale_direction(ramp->end_level,
 					   effect->direction);
 
-	*out_slope = abs(start_level - end_level) / 2;
-	*out_center = clamp_t(s32, (start_level + end_level) / 2,
+	*out_slope = DIV_ROUND_CLOSEST(abs(start_level - end_level), 2);
+	*out_center = clamp_t(s32, DIV_ROUND_CLOSEST(start_level + end_level, 2),
 			      S16_MIN, S16_MAX);
 
 	*out_invert = (start_level < end_level) ? 0x04 : 0x05;


Kimplul:
Thanks for the PR, I'm alright with the changes but I have a couple things I'd like to run by you before I merge:

> For small signed force values this can suppress effects and introduce a directional bias between positive and negative forces.

Does 'suppression' here mean that very weak effects don't really get played at all? I guess I can see how that would come about, some input scaled by the gain produces a small enough value that the wheel effectively just doesn't react as 'expected' (due to internal resistance or whatever). But I'm wondering what's the smallest value that the wheel reacts to, does this PR make it so that what was previously a 0 gain turns into 1 and now the wheel reacts, or something more like 15 to 16? (For future work, if it's the latter, should there be some kind of non-linear response where effects below the measured threshold get mapped above it?)

Or is 'suppressed' just that all effects are slightly weaker than they should be, it's just more noticeable with weak effects, in which case how does this differ from upping the gain a bit?

The symmetry stuff is interesting. Division itself always rounds towards zero, so isn't the operation itself symmetrical? Is the wheel's hardware itself just biased and applying slightly stronger effects hides this bias a bit better? There were apprently suprising asymmetries in the Windows driver, previously discussed [here](https://github.com/Kimplul/hid-tmff2/pull/105#issuecomment-2180277851), would this maybe be related?


> The change does not alter gain, range, device defaults or the T300 protocol layout. Intermediate calculations remain widened before division and the existing output limits are preserved.
> [...]
> Existing modpost warning about the composite module's MODULE_DESCRIPTION() remains unchanged.

Good to know, I guess? Haha

kakra:
Well, let me try to rephrase it differently:

In ETS2 I felt that effects have a tendency to apply stronger in one direction than the other. I first thought that this might happen because just cutting the integer instead of rounding it will cut it down to the next integer instead of towards zero.

From my tests with my own test program I know that weak effects do not overcome the wheel's internal friction but you can still feel their effect if applying a slight counter force, so motor vibrations from the truck can be felt even if the force is too weak to move the wheel. This vibration dominated in one direction while playing ETS2. Now with the rounding it feels more even.

Other games like FH5 do not show this because the strong forces used there hide the previously asymmetric weak effects.

I came to this idea because while developing xpadneo, I made pressure dependent trigger rumble to synthesize trigger vibrations from current input and game rumble. There I also added rounding to make the effect feel smoother, especially when gain was turned down. Additionally, for mouse pointer movement with the stick (xpadneo can emulate a mouse now), I used rounding and error diffusion and that made a big difference for pointer movement. Tho, error diffusion is not something we should apply to effects because we do not know when the next effects comes in. Error diffusion needs deterministic continuation of input value streams. This would prevent the wheel motors from completely shutting down when idle.

I've set gain to 75% (49151) as this is the default the T300 drivers ship with on Windows. I've set the other levels to 100%. For these patches, I think it only makes a difference during low spring forces and very subtle effects, like ETS2 uses.

> The symmetry stuff is interesting...

Yeah, I think integer division should always round down towards zero, but there could be cases where it does not. Usually, I'd say that C rounds toward zero, as per its spec. Python, for example, does not do this: It rounds down, so negative numbers become even smaller. In your code, I only found cases where division always works around a proper zero value (not shifted previously), and since we are in C, it means it will properly round towards zero aka the center point.

So maybe for ETS2 we are sitting just at a tipping point between rounding down or up, and now it properly rounds towards the nearest integer? This would mean that ETS2 itself is asymmetric which could actually be the case because it simulates engine torque: If you hit the throttle hard, one side of the truck will have its wheels pressed harder against the asphalt, and thus the forces on the steering become asymmetric, because the truck as a whole twists. The asymmetric effect is still there in ETS2 when throttling up, or climbing a steep hill. But it now feels more even when just cruising the highways.

I've been playing with this for multiple hours - and if it's not just imagination because I changed something, it felt different. ;-)

> ... all effects are slightly weaker than they should be ...

Kind of, maybe? I think if the effects cover enough range of values, this might not matter. But if a game is working with a lot fine details effects, the overall error by only rounding down becomes much bigger. E.g., if you always round 0.7 down to 0, that's a much bigger error (relatively speaking) than rounding down 5.7 to 5. Do this 60 times a second and the error adds up. I hope this explains what I think is happening.

> should there be some kind of non-linear response where effects below the measured threshold get mapped above it

I've been thinking about this, too. But some guy did some research by measuring the actual forces, and even the nativ Windows driver has a very flat force curve on low values. And from my own measurements by letting the wheel turn freely and sampling the input values in 5ms steps, the response looks almost identical with a tendency to be slightly weaker in Linux (still within a margin of error). I didn't compare it with the patches changes yet, tho, but the proper rounding may fix that.

So, I think, no: The Windows driver doesn't boost low forces.

But the Windows driver has a boost feature which just doubles some forces (not all), with the result of clipping high forces. Nobody should probably use that. But tmff2 could optionally implement that, of course.

Ref: https://www.briankoponen.com/thrustmaster-boost-force-feedback-analysis/

> Good to know, I guess? Haha

Yeah, this constantly pops up here. I didn't bother with it yet, doesn't seem harmful in any way. :-D

PS: The kernel patch checker also complains about missing module descriptions. But I'm ignoring that. That seems to be something that was added to later kernels because I got similar warnings in xpadneo at some point during development.

kakra:
> previously discussed [here](https://github.com/Kimplul/hid-tmff2/pull/105#issuecomment-2180277851), would this maybe be related?

Looks interesting. I can look into it and try to replicate the result using my setup. If you want, you can wait with the merge until then. My first intention was to get the gain identical first, so I can compare Windows and Proton. I usually don't play the Linux-native versions of games, just the Proton version. Under that assumption, I'm looking into getting the forces identical through that layer compared to native Windows.

I don't run Windows on my gaming rig for 15+ years now, so I don't have a side-by-side compare. I can only compare with my Windows laptop by running synthetic tests (applying effects, sample wheel movement).

Kimplul:
> So maybe for ETS2 we are sitting just at a tipping point between rounding down or up, and now it properly rounds towards the nearest integer? This would mean that ETS2 itself is asymmetric

So is this is at least partially game-specific? I guess that makes sense to me, games do their own calculations and could fairly easily land on a 'tipping point' by accident. Slightly surprised that such a small difference makes a noticeable difference, but I guess if integrated over time you'd get some real differences, at least if the wheel is rotating more-or-less freely.

Although,

> Other games like FH5 do not show this because the strong forces used there hide the previously asymmetric weak effects.

would imply that effects are universally asymmetric?

> Looks interesting. I can look into it and try to replicate the result using my setup. If you want, you can wait with the merge until then

Not a blocker, but would be neat to see the results. I'll try and make some time this evening to do some smaller things (udev/README/module description), I'll probably merge this PR while I'm at it.

kakra:
We can wait with the merge until I tested a few more things if you're more comfortable with that.

> > Other games like FH5 do not show this because the strong forces used there hide the previously asymmetric weak effects.
> 
> would imply that effects are universally asymmetric?

I don't think so. It just confirms that the accumulated error over an interval with stronger forces becomes negligible. Also FH5 probably doesn't use deliberate asymmetric effects. Still, the link you posted to the old discussion is worth some more investigation. The only thing I noticed in FH5 is that if you rev the engine, the car will slightly roll. But this is probably purely a cosmetic effect, nothing relevant physically for force feedback. That is different in ETS2: You can feel the engine torque in the steering wheel. FH5 seems to be much simpler, it even uses a weak auto center effect, even when the wheels stand still and thus should physically prevent such an effect. I think I should try a few games with more realistic physics than FH5.

About the asymmetric effects, this might be a more recent addition to ETS2 when they reworked truck physics. I don't let the wheel run freely: While cruising just straight roads, I usually hold the wheel with like just two fingers, IOW, very loosely. ETS2 always had a tendency to steer slightly right due to engine torque simulation. This is a deliberate effect in ETS2. But now with the patches, it works slightly different: The effect is still there but only if you hear the engine starting to work harder, e.g. when climbing up a slope.

It is probably important to adjust the in-game FFB the way I did. I followed recommendations from this page which seems to compare and test a lot of different combinations: https://www.briankoponen.com/euro-truck-simulator-2-thrustmaster-tx-t300-settings/

It is probably also important to run with fuzz=0 and flat=0 (which are the new defaults since the previous PR you merged), so tiny adjustments actually reflect into the game.

BTW: Getting back to a previous discussion, I found out that FH5 even on native Windows sends inverted forces to the T300 although it has a built-in support profile for it which is properly detected and applied in Proton. The game itself probably has a setting to invert forces for that very reason. So there is actually no problem with inverted forces in your driver, just flipping the setting will fix the game. But there's still a pedal ordering problem, but that bug is in wine alone, the Linux evdev interface is correct.

Kimplul:
I went ahead and merged, since even with the asymmetry mystery, rounding to the closest integer probably makes more sense overall. If you do look further into the Windows driver asymmetries, please do report back if you happen to figure something out, I'm rather curious what that's all about and haven't really come up with any sensible explanations myself.

Anyway, thanks a lot!

kakra:
Thanks for merging, @Kimplul. I just wondered if we should also change:
```c
fixp_sin16(direction * 360 / 0x10000)
```
to proper rounding. Currently, the formula relies on floor rounding to scale 0..65535 to 360 degrees. It's technically correct because that's how you scale from one zero-based range to another. But it doesn't round properly, seemingly to avoid wrap around to 360 degrees. I think we should convert it to `fixp_sin32_rad()` to get more angular precision  in the first place: the direction value is sub-degree which the current formula cuts to discrete degree steps. After all, setting fuzz=0 in the previous merge already showed that even 1 degree is a lot on my wheel (fuzz made up about 3 degrees which seems tiny but really is huge).

I'll send another PR for it, unless you'd like to quickly change it yourself:
```c
static s32 t300rs_scale_direction(s16 level, u16 direction)
{
	/* The Linux FF direction is a 16-bit turn. Keep its sub-degree
	 * precision instead of reducing it to an integer degree first. */
	return DIV_S64_ROUND_CLOSEST((s64)level *
		fixp_sin32_rad(direction, 0x10000), 0x7fffffff);
}
```

I'll test this local change for now, and report back. If you are not faster than me, I'll send a PR. ;-)

BTW: Before blindly taking this, it may need reconsideration of the changes from #105.

Kimplul:
> I'll test this local change for now, and report back. If you are not faster than me, I'll send a PR. ;-)

I am nothing if not glacially slow, hah. Feel free to open up a PR.

> BTW: Before blindly taking this, it may need reconsideration of the changes from https://github.com/Kimplul/hid-tmff2/pull/105.

Sorry, how do you mean?  Just that the higher resolution might produce values outside of the expected range or something else?

kakra:
> Sorry, how do you mean? Just that the higher resolution might produce values outside of the expected range or something else?

No, due to the better resolution, the effects of rounding down that previously limited the formula to only 360 distinct values (shifted to the lower than intended value), become low enough to reverse the effects of observed asymmetry which that PR probably tried to compensate. The difference between the absolute lower and upper bound can be explained by measuring exactly 1 degree of the total range: If you always round down to 360 distinct values, you get exactly that imbalance of 91 units: 65536 divided by 360 degrees make 182 units, statistically half of a game's intended forces are above, the other half is below the applied force, thus it's off by 91 units on average, which explains the off-center or asymmetric observation. Replacing the coarse 360 degrees model with a 16 bit unit will most likely eliminate that.

E.g., you wrote:

> 32677 is a bit surprising, I would've expected it to be 32768.

And @MmAaXx500 replied:

> I think it maps to some lookup table inside the wheel, although the value choice still odd, but it works out perfectly.

I'd argue that the difference of 91 can be perfectly explained from the error of always rounding down when scaling down to only 360 distinct degrees.

My suggestion from above should only be off by an average error of 1 or 2 (due to scaling from -65536..65536 on Windows to -32768..32767 on Linux).

So, the phase value mapping should be re-evaluated before merging my proposed change.

PR #215, opened 2026-08-22
Add support for the Thrustmaster T-GT II

### LINKED ISSUE #151: T-GT II - Enchantments

ranjian0:
The T-GT II ships in a T300 compat mode and shares the USB product ID `b66d`
with the T300RS, reporting a distinct `bcdDevice` of `0x0700

Fixes #151

### DIFF
--- a/Kbuild
+++ b/Kbuild
@@ -5,4 +5,5 @@ hid-tmff-new-y := \
 		src/tmt248/hid-tmt248.o \
 		src/tmtx/hid-tmtx.o \
 		src/tmtsxw/hid-tmtsxw.o \
-		src/tmtspc/hid-tmtspc.o
+		src/tmtspc/hid-tmtspc.o \
+		src/tgt2/hid-tgt2.o
--- a/README.md
+++ b/README.md
@@ -196,5 +196,10 @@ for wheels:
   and add `options hid-tmff-new timer_msecs=NUMBER` into it.
   The default timer period is 8, but numbers as low as 2 should work alright.
 
-+ The T-GT II might show up as a T300 at the moment, since it reuses the T300
-  USB product ID.
++ The T-GT II reuses the T300 USB product ID. Since the two wheels are not
+  distinguishable at probe time, T-GT II support (rotary encoders, ministicks,
+  analog L2/R2 and the correct pedal order) must be enabled explicitly with
+  the `tgt2_force=1` module parameter, e.g. in
+  `/etc/modprobe.d/hid-tmff-new.conf`:
+  `options hid-tmff-new tgt2_force=1`.
+  Without it, the wheel is handled as a T300.
--- a/src/hid-tmff2.c
+++ b/src/hid-tmff2.c
@@ -48,6 +48,12 @@ module_param(gain, ushort, 0);
 MODULE_PARM_DESC(gain,
 		"Level of gain (0-65535)");
 
+static bool tgt2_force;
+module_param(tgt2_force, bool, 0444);
+MODULE_PARM_DESC(tgt2_force,
+		"Treat 044f:b66d as a T-GT II instead of a T300RS, since the two "
+		"wheels share the same USB product ID");
+
 static u16 tmff2_scale_gain(u16 value)
 {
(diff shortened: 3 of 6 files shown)


Kimplul:
Thanks for the PR, but I vehemently disagree with setting a module parameter to detect which wheel we're talking to. There's some indication that checking the `bcdDevice` field can be used to separate wheels that show up with the same ID, see https://github.com/Kimplul/hid-tmff2/blob/78784253d0c5b9bdc04a7a9b7e3eddda55604de2/src/hid-tmff2.c#L695

Did you capture the `rdesc` from Windows, or what's this about 'cannot be presented as axes'? I'd like to avoid handling raw events as far as possible. I'd recommend capturing the `rdesc` on Windows, apprently the Windows driver modifies the `rdesc` sent over by the wheel quite heavily (kind of just replaces it completely, hah) and checking how the rotary encoders are represented there.

ranjian0:
> Thanks for the PR, but I vehemently disagree with setting a module parameter to detect which wheel we're talking to. There's some indication that checking the `bcdDevice` field can be used to separate wheels that show up with the same ID, see
> 
> https://github.com/Kimplul/hid-tmff2/blob/78784253d0c5b9bdc04a7a9b7e3eddda55604de2/src/hid-tmff2.c#L695
> 
> Did you capture the `rdesc` from Windows, or what's this about 'cannot be presented as axes'? I'd like to avoid handling raw events as far as possible. I'd recommend capturing the `rdesc` on Windows, apprently the Windows driver modifies the `rdesc` sent over by the wheel quite heavily (kind of just replaces it completely, hah) and checking how the rotary encoders are represented there.

1. I did not capture rdesc from windows. I used usbhid-dump with the driver unbound.
2. A module parameter is indeed ugly, I was not sure the best way to handle this since I only have one wheel to test.

I will do a capture from windows to see how the rotary encoders are mapped there

ranjian0:
- Managed to capture on Windows, and the wheel's own descriptor declares everything properly, so raw_event is not needed
- bcdDevice is precisely what I needed; mode handling has been implemented in the hid-tminit repo based on existing conventions.

Hopefully this is a much cleaner implementation.

Kimplul:
Thanks a lot! This indeed looks much better. Interesting that `0xb66d` is used for the device when it's still in the 'bootloading' phase, but I guess that's not totally unexpected, Thrustmaster seems to reuse IDs a bit willy-nilly.

I'm somewhat intrigued by `tgt2_calibrate()`. There are some similarities with https://github.com/Kimplul/hid-tminit/blob/8c4547288a6c182ed4ff131e36f710f11a76c4a9/tminit.c#L27, but apparently done here *after* initialization. Weird. The comments mention copying those packets from the windows captures, but did you try to omit `tgt2_calibrate()` (except for the packet for 'opening' the wheel) and what happened then?

I also don't entirely understand why the `open` packets are sent out during calibration and not with the explicit `open`/`close` callbacks.

ranjian0:
> Thanks a lot! This indeed looks much better. Interesting that `0xb66d` is used for the device when it's still in the 'bootloading' phase, but I guess that's not totally unexpected, Thrustmaster seems to reuse IDs a bit willy-nilly.
> 
> I'm somewhat intrigued by `tgt2_calibrate()`. There are some similarities with https://github.com/Kimplul/hid-tminit/blob/8c4547288a6c182ed4ff131e36f710f11a76c4a9/tminit.c#L27, but apparently done here _after_ initialization. Weird. The comments mention copying those packets from the windows captures, but did you try to omit `tgt2_calibrate()` (except for the packet for 'opening' the wheel) and what happened then?
> 
> I also don't entirely understand why the `open` packets are sent out during calibration and not with the explicit `open`/`close` callbacks.

You made some good points. I am currently running some experiments to clean up the tgt2_calibrate and use the open callback. It's taking a bit longer than expected, but I will submit new changes when everything is tested.

PR #218, opened 2026-08-28
T500RS support

cazzoo:
New version of the T500RS driver supporting all effects and all games


cazzoo:
@hoover67, @thefahre, I updated the codebase for the driver from your captures and ended-up with a state that should be much better. Please let me know your thoughts

hoover67:
Hey @cazzoo thanks for all the updates. What's the current procedure to check out the driver source including your commits? I'm a bit at a loss on how to do this TBH. 

All the best, 

Uwe

thefahre:
Hi @cazzoo, I gave it a spin, and FFB does work, but it feels like all the force feedback is compressed. Small force got very exaggerated in center; wheel tries way too hard to get back to center, and turning makes all the extra forces disappear and clips. 

Settings on oversteer and the game is attached:

Screenshot_20260829_190507
Screenshot_20260829_190542

src/hid-tmff2.c
kakra, line 1: @Kimplul I think this should go into the master branch separately.

udev/99-thrustmaster.rules
kakra, line 1: @Kimplul This should go into the master branch separately, but along with all supported models having fuzz/flat set to 0. I think my previous patch may only have adjusting this for T300 models.

kakra:
I think the rework towards workqueue-based effects should go into the master branch first as a separate, clean PR, with testing results and semantically distinct, bisectable commits. Also, a later commit should not fix a bug that a previous commit introduced. Rather fold them, and/or use "fixup!" prefixes to mark a fixing commit for later squashing.

To get this merged, I'd suggest to make it easier to review:

1. Split your series into bug fixes, that should go in first, then model-agnostic architecture preparation, foundational work, and finally the driver code itself.
2. Use focused, semantically distinct, bisectable commits with proper reasoning and explanation in each commit message.
3. Avoid broad style rewrites across the driver base, follow existing code style patterns.
4. Use a pre-commit hook to prevent code with linting or clang format errors go into a commit in the first place instead of adding a commit later which fixes these touching multiple previous committed lines.
5. Follow the maintainers commit message habits.

All these semantic commits can live in one single draft PR, even out of order. If something works and is ready, cherry-pick into a new branch in proper order, run tests on each commit, fix remaining bugs, create a PR, then rebase your queued commits on top of the split-out series.

udev/99-thrustmaster.rules
Kimplul: I think this diff just reinserts the file I previously removed, with the T500 added. As such, I don't think it should be included at all. Your patch should apply to all wheels added to this driver, IIUC.

kakra: > I think this diff just reinserts the file I previously removed

Okay, yeah, the PR is a bit messy, and I outlined below how to clean it up.

> Your patch should apply to all wheels added to this driver

Yes, maybe: I think a first version only touched the T300, any I remember you asked me to apply to just every model. ;-)

src/hid-tmff2.c
```diff
@@ -248,9 +294,38 @@ static ssize_t gain_store(struct device *dev,
 	}
 
 	gain = value;
-	if (tmff2->set_gain) /* if we can, update gain immediately */
-		tmff2->set_gain(tmff2->data, gain);
 
+	if (!tmff2->set_gain)
+		return count;
+
+	/* Rationale: two-level gain model
```
Kimplul, line 301: I looked back at the previous PRs and IIUC, the original reason for inserting `set_gain` etc. into the workqueue was that you encountered a crash if it was done directly.  I had a quick look and found at least one other driver that sends out USB messages directly in the attribute handler: https://elixir.bootlin.com/linux/v7.2.2/source/drivers/hid/hid-appletb-kbd.c#L126 (where `appletb_kbd_set_mode` ultimately calls `hid_hw_request`).

That would point towards something being buggy, either that other driver, the USB layer (when dealing with the T500?) or some previous development version of this PR that these changes try to work around.

I'm alright with workarounds, but I'd like to know what we're working around, could you try to check that out?

udev/99-thrustmaster.rules
cazzoo: I removed the file from my code as well, as per the base driver

cazzoo:
> Hi @cazzoo, I gave it a spin, and FFB does work, but it feels like all the force feedback is compressed. Small force got very exaggerated in center; wheel tries way too hard to get back to center, and turning makes all the extra forces disappear and clips.
> 
> Settings on oversteer and the game is attached:

Can you please run another session with the following:
- Run AC, then: dmesg -w | grep -i autocenter → the driver's warning appears the moment the game initializes FFB or not.
- Or run with Oversteer open during play (it re-stomps autocenter to 0) → check if forces suddenly feel normal

thefahre:
> > Hi @cazzoo, I gave it a spin, and FFB does work, but it feels like all the force feedback is compressed. Small force got very exaggerated in center; wheel tries way too hard to get back to center, and turning makes all the extra forces disappear and clips.
> > Settings on oversteer and the game is attached:
> 
> Can you please run another session with the following:
> 
> * Run AC, then: dmesg -w | grep -i autocenter → the driver's warning appears the moment the game initializes FFB or not.
> * Or run with Oversteer open during play (it re-stomps autocenter to 0) → check if forces suddenly feel normal

Hi @cazzoo, 

- Run AC, then: dmesg -w | grep -i autocenter → the driver's warning appears the moment the game initializes FFB or not: [  104.772965] tmff2 0003:044F:B65E.000F: Game might have set autocenter to 100%, you might want to set it back to expected value using oversteer (or keep oversteer open) or system gain.
- Or run with Oversteer open during play (it re-stomps autocenter to 0) → check if forces suddenly feel normal: running Oversteer with root permission, no change, FFB still feels very off.

cazzoo:
closing this one (will use stacked PRs)

PR #219, opened 2026-09-09
Feature/t500rs driver only

cazzoo:


Kimplul:
Heyo, what's your plan with this draft? Looks like the changes are more or less the same as in https://github.com/Kimplul/hid-tmff2/pull/218 but I guess the commit history is more squashed as suggested in https://github.com/Kimplul/hid-tmff2/pull/218#issuecomment-5467923695. If you tried to follow the suggestions in https://github.com/Kimplul/hid-tmff2/pull/218#issuecomment-5467923695, this PR kind of misses the mark a bit. For example, https://github.com/Kimplul/hid-tmff2/pull/219/commits/ddc030525ed76b09a9e5499bf972a1d5f2801b9f could've/should've been a preparatory PR, since it's more generic and touches other things than just the T500.

It was kind of implied, but maybe not explicitly stated, but size of this PR is kind of the main issue here. The last few PRs had hundreds of comments and I can see that at least some of them aren't reflected in this version (which is understandable, there's a huge amount of context to keep track of).

GitHub has a new feature, stacked PRs, that you can use to create a series of PRs that each depend on eachother. That way, you can split the total work into smaller bits that can be reviewed separately, and each thread is more focused. For example, for this PR something like

1. Defer gain/autocenter (generic changes, required for the T500 to work)
3. T500RS with just the constant effects supported, including docs for just the constant effect (most games just use constant effects so this is already an excellent addition to this driver!)
4. Periodic, ramp effects (+ docs for them)
5. Passive effects (spring, damper) (+ docs for them)

or something along those lines.

[This](https://graphite.com/blog/code-review-best-practices) is a pretty good overview of what a good PR consists of. Note that it suggests keeping PRs under 400 lines of code, compared to the 2900 in this one :) Although, please don't take the 400 lines as a hard limit, I would imagine that the first T500 commit has to be a bit bigger since there's some boilerplate etc. that needs to exist.

cazzoo:
Hi @Kimplul,

I am indeed in the works to split in chunks the driver.
I just moved the gain/autocenter defer onto a dedicated PR that I intent to open at first.
Then, for this one I was intending to have only 1 for the whole driver, but if I need to split it as well in small chunks, I can do it.

I will thy that "PR stack" feature github provides, hopefully it will make it clearer to review. Will cancel the previous ones

PR #220, opened 2026-09-09
tmff2: defer gain and autocenter changes to the workqueue

cazzoo:
The input core invokes the FF gain and autocenter callbacks from event context, which may run with IRQs disabled, and the sysfs gain store calls them from process context. Model drivers issue blocking HID output in these callbacks, which is illegal in atomic context and races the effect worker on shared transfer buffers.

Route both through the existing effect workqueue instead: store the latest request in the device entry and apply it in the worker, serialized with effect processing. The sysfs gain store re-queues the last in-game gain so level changes take effect immediately; in-game gain continues to be scaled by the sysfs gain parameter via tmff2_scale_gain(), now applied in the worker.

### DIFF
--- a/src/hid-tmff2.c
+++ b/src/hid-tmff2.c
@@ -235,6 +235,7 @@ static ssize_t gain_store(struct device *dev,
 		struct device_attribute *attr, const char *buf, size_t count)
 {
 	struct tmff2_device_entry *tmff2 = tmff2_from_hdev(to_hid_device(dev));
+	unsigned long flags;
 	u16 value;
 	int ret;
 
@@ -248,8 +249,15 @@ static ssize_t gain_store(struct device *dev,
 	}
 
 	gain = value;
-	if (tmff2->set_gain) /* if we can, update gain immediately */
-		tmff2->set_gain(tmff2->data, gain);
+
+	/* Mark the gain dirty; the worker combines this master gain with
+	 * the last input gain and sends the product to the wheel */
+	spin_lock_irqsave(&tmff2->lock, flags);
+	__set_bit(FF_EFFECT_QUEUE_GAIN, &tmff2->pending_flags);
+	spin_unlock_irqrestore(&tmff2->lock, flags);
+
+	if (!delayed_work_pending(&tmff2->work) && tmff2->allow_scheduling)
+		schedule_delayed_work(&tmff2->work, 0);
 
 	return count;
 }
@@ -264,6 +272,7 @@ static DEVICE_ATTR_RW(gain);
 static void tmff2_set_gain(struct input_dev *dev, uint16_t value)
 {
 	struct tmff2_device_entry *tmff2 = tmff2_from_input(dev);
+	unsigned long flags;
 
 	if (!tmff2)
 		return;
@@ -273,13 +282,21 @@ static void tmff2_set_gain(struct input_dev *dev, uint16_t value)
 		return;
 	}
 
-	if (tmff2->set_gain(tmff2->data, tmff2_scale_gain(value)))
-		hid_warn(tmff2->hdev, "unable to set gain\n");
+	/* Defer to workqueue: store the game gain and schedule. The worker
+	 * scales it with the sysfs gain via tmff2_scale_gain() */
+	spin_lock_irqsave(&tmff2->lock, flags);
+	tmff2->input_gain = value;
+	__set_bit(FF_EFFECT_QUEUE_GAIN, &tmff2->pending_flags);
+	spin_unlock_irqrestore(&tmff2->lock, flags);
+
+	if (!delayed_work_pending(&tmff2->work) && tmff2->allow_scheduling)
+		schedule_delayed_work(&tmff2->work, 0);
 }
 
 static void tmff2_set_autocenter(struct input_dev *dev, uint16_t value)
 {
 	struct tmff2_device_entry *tmff2 = tmff2_from_input(dev);
+	unsigned long flags;
 
 	if (!tmff2)
 		return;
@@ -289,8 +306,14 @@ static void tmff2_set_autocenter(struct input_dev *dev, uint16_t value)
 		return;
 	}
 
-	if (tmff2->set_autocenter(tmff2->data, value))
-		hid_warn(tmff2->hdev, "unable to set autocenter\n");
+	/* Defer to workqueue: store pending autocenter and schedule */
+	spin_lock_irqsave(&tmff2->lock, flags);
+	tmff2->pending_autocenter = value;
+	__set_bit(FF_EFFECT_QUEUE_AUTOCENTER, &tmff2->pending_flags);
+	spin_unlock_irqrestore(&tmff2->lock, flags);
+
+	if (!delayed_work_pending(&tmff2->work) && tmff2->allow_scheduling)
+		schedule_delayed_work(&tmff2->work, 0);
 }
 
 static void tmff2_work_handler(struct work_struct *w)
@@ -303,9 +326,34 @@ static void tmff2_work_handler(struct work_struct *w)
 	unsigned long time_now;
 	__u16 effect_delay, effect_length;
 
+	uint16_t pending_gain = 0, pending_autocenter = 0;
+	bool set_gain = 0, set_autocenter = 0;
 
 	if (!tmff2)
 		return;
+
+	/* Apply pending control changes (gain/autocenter) in process context */
+	spin_lock_irqsave(&tmff2->lock, lock_flags);
+
+	if (test_bit(FF_EFFECT_QUEUE_GAIN, &tmff2->pending_flags)) {
+		pending_gain = tmff2->input_gain;
+		__clear_bit(FF_EFFECT_QUEUE_GAIN, &tmff2->pending_flags);
+		set_gain = 1;
+	}
+
+	if (test_bit(FF_EFFECT_QUEUE_AUTOCENTER, &tmff2->pending_flags)) {
+		pending_autocenter = tmff2->pending_autocenter;
+		__clear_bit(FF_EFFECT_QUEUE_AUTOCENTER, &tmff2->pending_flags);
+		set_autocenter = 1;
+	}
+
+	spin_unlock_irqrestore(&tmff2->lock, lock_flags);
+
+	if (set_gain && tmff2->set_gain)
+		tmff2->set_gain(tmff2->data, tmff2_scale_gain(pending_gain));
+
+	if (set_autocenter && tmff2->set_autocenter)
+		tmff2->set_autocenter(tmff2->data, pending_autocenter);
 
 	for (effect_id = 0; effect_id < tmff2->max_effects; ++effect_id) {
 		unsigned long actions = 0;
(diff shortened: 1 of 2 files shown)


Kimplul:
Thanks! I'll copy a comment I previously left on this matter:

I looked back at the previous PRs and IIUC, the original reason for inserting set_gain etc. into the workqueue was that you encountered a crash if it was done directly. I had a quick look and found at least one other driver that sends out USB messages directly in the attribute handler: https://elixir.bootlin.com/linux/v7.2.2/source/drivers/hid/hid-appletb-kbd.c#L126 (where appletb_kbd_set_mode ultimately calls hid_hw_request).

That would point towards something being buggy, either that other driver, the USB layer (when dealing with the T500?) or some previous development version of this PR that these changes try to work around.

I'm alright with workarounds, but I'd like to know what we're working around, could you try to check that out?

src/hid-tmff2.c
```diff
@@ -248,8 +249,16 @@ static ssize_t gain_store(struct device *dev,
 	}
 
 	gain = value;
-	if (tmff2->set_gain) /* if we can, update gain immediately */
-		tmff2->set_gain(tmff2->data, gain);
+
+	/* Re-queue the last in-game gain; the worker scales it with the
+	 * new sysfs gain via tmff2_scale_gain() */
+	spin_lock_irqsave(&tmff2->lock, flags);
+	tmff2->pending_gain = tmff2->last_input_gain;
```
Kimplul: Are there 'two' different changes in this commit? The first being moving `set_gain` to the workqueue, and the other seemingly fixes a bug where setting the gain via `sysfs` overrides the in-game requested gain?

I added a comment about the first, and I think this approach to the second one is a bit messy. I wonder if it might not make more sense to have `sysfs` send over the gain to the wheel, and `tmff2_set_gain` then sets a game-specific gain that's just a software calculation, probably something like `t300rs_calculate_constant_level(level, gain, direction)`. Although that still leaves a situation like `Game1` requests 50% gain, user exits, `Game2` doesn't set a gain level. Now changing the `sysfs` attribute doesn't increase `Game2`'s gain beyond 50%. Maybe resetting the game-specific gain if a game exits, but not sure how that would be monitored. `close` doesn't seem to be called for situations like the above, not entirely sure why. Maybe `evdev` keeps an open FD or something?

src/hid-tmff2.h
```diff
@@ -93,8 +104,8 @@ struct tmff2_device_entry {
 	ssize_t (*alt_mode_show)(void *data, char *buf);
 	ssize_t (*alt_mode_store)(void *data, const char *buf, size_t count);
 	int (*set_autocenter)(void *data, uint16_t autocenter);
-	__u8 *(*wheel_fixup)(struct hid_device *hdev, __u8 *rdesc, unsigned int *rsize);
```
Kimplul, line 96: Some kind of whitespace change, remove?

src/hid-tmff2.c
Kimplul: Come to think of it, on Windows there's a button that can select between letting the game or the Thrustmaster wheel manager set the gain. `sysfs` is kind of like the wheel manager, and `input` is the game, and I wonder if the button just ignores which source is used as the 'actual' gain value.

Should probably check with something like `lg4ff` what they do in this situation.

Kimplul: `new-lg4ff` doesn't seem to have a separate wheel command for gain, but yeah, it has a separate `master_gain` for `sysfs` gain and a game-specific `gain` that are used to calculate the final gain that is sent over to the wheel. Presumably the scenario of `Game2` not setting a gain is rare enough as to not really be an issue?

Kimplul:
> Model drivers issue blocking HID output in these callbacks

Ah, I think I see what's going on here. The next PR uses `hid_hw_output_report()` while the T300 uses `hid_hw_request()`. Presumably `hid_hw_request()` is fine to use in this context. Why does the T500 not use `hid_hw_request()`?

cazzoo:
> Thanks! I'll copy a comment I previously left on this matter:
> 
> I looked back at the previous PRs and IIUC, the original reason for inserting set_gain etc. into the workqueue was that you encountered a crash if it was done directly. I had a quick look and found at least one other driver that sends out USB messages directly in the attribute handler: https://elixir.bootlin.com/linux/v7.2.2/source/drivers/hid/hid-appletb-kbd.c#L126 (where appletb_kbd_set_mode ultimately calls hid_hw_request).
> 
> That would point towards something being buggy, either that other driver, the USB layer (when dealing with the T500?) or some previous development version of this PR that these changes try to work around.
> 
> I'm alright with workarounds, but I'd like to know what we're working around, could you try to check that out?

Dug into it — here's the concrete path. The freeze (whole system hang, nothing in dmesg) happened with the T500 whenever gain/autocenter changed while effects were running, e.g. dragging Oversteer's sliders.

The problematic context is not sysfs — the sysfs store runs in process context and may sleep, which is why appletb-kbd doing hid_hw_request() from an attribute handler is fine. The crash path is the input event path: userspace writes an EV_FF/FF_GAIN event → input_inject_event() takes dev->event_lock with IRQs disabled (input.c (https://elixir.bootlin.com/linux/v6.18/source/drivers/input/input.c#L445), guard(spinlock_irqsave)) → input_handle_event() → dev->event() which is input_ff_event() (https://elixir.bootlin.com/linux/v6.18/source/drivers/input/ff-core.c#L193) → ff->set_gain() → our driver callback issues a synchronous USB transfer, i.e. sleeps under a spinlock with IRQs off. That's the freeze.

Note the T300 path has the same latent contract violation on master today (t300rs_set_gain → hid_hw_request() → usb_control_msg(), also a sleeping call) — it just rarely bites since gain/autocenter changes are infrequent. Deferring both callbacks to the existing effect worker fixes the contract for every wheel model, not just the T500.

cazzoo:
> > Model drivers issue blocking HID output in these callbacks
> 
> Ah, I think I see what's going on here. The next PR uses `hid_hw_output_report()` while the T300 uses `hid_hw_request()`. Presumably `hid_hw_request()` is fine to use in this context. Why does the T500 not use `hid_hw_request()`?

Two reasons:

1. It wouldn't remove the need for the deferral. hid_hw_request() bottoms out in usb_control_msg() so it's exactly as illegal as hid_hw_output_report() under event_lock with IRQs off. The appletb-kbd comparison works because its handler runs from sysfs (process context).
2. Protocol fidelity. The T500RS FFB protocol (from Windows USB captures) is a stream of fixed-layout output reports on the interrupt-OUT pipe (0x42 init, 0x01/0x02/0x03 effect packets, 0x41 START/STOP, 0x43 gain…). The driver sends those byte sequences verbatim via hid_hw_output_report(). Whether this firmware accepts the same payloads as SET_REPORT control transfers is untested, and I'd rather keep the wire traffic byte-identical to the captures than change transport to dodge a locking rule the deferral already satisfies.

Once the gain/autocenter HID I/O runs from the worker, either transport would be legal — so if you prefer hid_hw_request() for consistency with the T300, that's testable follow-up work on the T500 PR. The deferral is needed either way.

src/hid-tmff2.c
cazzoo: > `new-lg4ff` doesn't seem to have a separate wheel command for gain, but yeah, it has a separate `master_gain` for `sysfs` gain and a game-specific `gain` that are used to calculate the final gain that is sent over to the wheel. Presumably the scenario of `Game2` not setting a gain is rare enough as to not really be an issue?

Fair point! I force-pushed a rework that removes the last_input_gain re-apply hack entirely. Gain is now plain live state:

- tmff2_set_gain (input) stores the game gain and marks the gain pending
- the sysfs store (master) just marks the gain pending
- the worker computes the product master × game via tmff2_scale_gain() at flush time and sends one hardware gain command

So a sysfs change still takes effect immediately, but there's no replay logic. The "second change" is gone and the commit is one mechanism. (The inline comments may show as outdated after the force-push; current code is at gain_store/tmff2_set_gain/tmff2_work_handler.)

On the lg4ff comparison, that's exactly the model now implemented: master_gain (sysfs) × game gain (input) → final gain sent to the wheel, matching new-lg4ff. 
One difference though: the T300/T500 do have a hardware gain command, so the product goes over the wire as the wheel gain rather than being applied to effect levels in software. This also keeps behaviour for existing wheels identical to what the worker already sent when a game set gain.

The "Game2 never sets a gain" edge case. I agreed it's rare enough to accept (same assumption as lg4ff). The game gain persists per-device until the next input write. Detecting game exit isn't reliable here for the reason you mentioned

Kimplul:
> Dug into it — here's the concrete path. The freeze (whole system hang, nothing in dmesg) happened with the T500 whenever gain/autocenter changed while effects were running, e.g. dragging Oversteer's sliders.
>
> The problematic context is not sysfs — the sysfs store runs in process context and may sleep, which is why appletb-kbd doing hid_hw_request() from an attribute handler is fine. The crash path is the input event path: userspace writes an EV_FF/FF_GAIN event → input_inject_event() takes dev->event_lock with IRQs disabled (input.c (https://elixir.bootlin.com/linux/v6.18/source/drivers/input/input.c#L445), guard(spinlock_irqsave)) → input_handle_event() → dev->event() which is input_ff_event() (https://elixir.bootlin.com/linux/v6.18/source/drivers/input/ff-core.c#L193) → ff->set_gain() → our driver callback issues a synchronous USB transfer, i.e. sleeps under a spinlock with IRQs off. That's the freeze.

[Please don't be a meat proxy.](https://gruhn.me/blog/2026-08-03/). The above links to source code don't correspond to the supposed execution path, and the given cause for the freeze, dragging Oversteer sliders, uses the `sysfs` attribute, but

> The problematic context is not sysfs

which is contradictory, no?

Kimplul:
> It wouldn't remove the need for the deferral. hid_hw_request() bottoms out in usb_control_msg() so it's exactly as illegal as hid_hw_output_report() under event_lock with IRQs off. The appletb-kbd comparison works because its handler runs from sysfs (process context).

I couldn't find the path to `usb_control_msg()`. As far as I can tell, the path ends up at https://elixir.bootlin.com/linux/v7.2.2/source/drivers/hid/usbhid/hid-core.c#L525, which inserts requests into a queue. I also found some other drivers sending HID requests with `hid_hw_requesr()` in the `set_gain` callback, I guess most notably https://elixir.bootlin.com/linux/v7.2.2/source/drivers/hid/usbhid/hid-pidff.c#L700. The above would imply those drivers are buggy as well, IIUC, which is not impossible but I'm skeptical.

If we go with the approach of just setting an internal `game_gain` (or something along those lines) variable in `set_gain`, the request is skipped altogether, and since the `sysfs` attribute is in user context, we wouldn't need the deferral mechanism in any case, buggy or not.

For what it's worth, I can see how it's nice to not have to rely on external functions being atomic-safe, especially when they're not documented for it.
It's just that it seems we can get away with something easier to understand here. Moving the incoming requests to a separate workqueue causes the control flow to jump abruptly, making it difficult to see what happens next, which I'd like to avoid if possible.

As a bit of history, the workqueue was originally just added to limit how often the driver sends out requests to the wheel, not to avoid atomic contexts. Adding a second responsibility arguably muddles the waters a bit from a perspective of the mental model needed to understand this driver.

PR #221, opened 2026-09-10
hid-tmff2: add T500RS wheel support (constant force)

cazzoo:
Stacked on #220 (defer gain/autocenter to the workqueue).

Adds a T500RS driver on top of the core hid-tmff2 framework:

- HID-mode init sequence (0x42 sync, slot STOP sweep, 0x40 autocenter teardown, 0x43 gain default); the b65d bootloader -> b65e mode switch is handled by hid-tminit.
- Constant force effects only at this stage: hardware slot 0 with fixed subtypes (param 0x0e / env 0x1c), 0x03 level updates, mandatory zero-envelope 0x02, and a software expiry timer sending per-slot 0x41 STOPs (the wheel never auto-stops on its own).
- Gain and range via the core parameters; autocentering routed through the core workqueue (T500RS callbacks do blocking HID I/O and must not run from atomic context).
- Protocol reference: docs/T500RS_FFBEFFECTS.md (constant force sections). The shared docs/FFBEFFECTS.md is renamed to docs/T300RS_FFBEFFECTS.md since it documents the T300RS protocol.
- udev rule for the SteamDeck variant.

Follow-up stages in this stack add periodic/ramp effects and passive (condition) effects.


Kimplul:
> Gain and range via the core parameters; autocentering routed through the core workqueue (T500RS callbacks do blocking HID I/O and must not run from atomic context).

This bit is the preparatory PR, and shouldn't really be included in this PR. You should be able to change the target of this PR to #220 instead of `master`, which splits off the commits that are not about the T500 constant force changes. That's the idea of stacked PRs anyway.

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,239 @@
...
+
+The `effect_id` byte mirrors the **hardware slot** the effect runs on - it is
+not always `0x00`:
+
+- Constant force runs on **slot 0** -> `effect_id = 0x00`.
+- Other effect families (added in later stages of this driver) get slots
+  `1, 2, 3, ...` assigned sequentially -> their `0x01` uploads and `0x41`
+  START/STOP packets carry that slot number (`41 01 41 ff` starts slot 1).
+
+### Why the wheel never stops on its own
```
Kimplul, line 62: This was discussed previously, packet uploads have a `duration` field which presumably tells the wheel how long an effect should run. This contradicts that, no?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+/* Supported parameters */
+static unsigned long t500rs_params = PARAM_GAIN | PARAM_RANGE;
+
+/* Supported effects. */
+const signed short t500rs_effects[] = { FF_CONSTANT,
+					FF_GAIN,       FF_AUTOCENTER,
+					-1 };
+
+/*
+ * Resolve the hardware effect slot index for a given effect.
```
Kimplul, line 327: This is new, didn't we last time conclude that `effect_id + 1` was the way to go?

docs/T500RS_FFBEFFECTS.md
```diff
@@ -0,0 +1,239 @@
...
+
+| Code | Effect                        |
+|------|-------------------------------|
+| 0x00 | Constant force                |
+
+**Duration note:** for constant effects the wheel ignores the duration and runs
+until stopped, so the driver sends `0xffff` ("infinite") and relies on its own
+timer (paragraph 3).
+
+### 5.2 Envelope - `0x02` (9 bytes)
```
Kimplul, line 123: If you'd like, envelopes can be split off into a separate PR as well, since they're shared across several different effects. Documenting them here and then just saying "they're not used" feels a bit weird, especially since they're apparently a separate packet?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+
+	/* Validate effect ID range */
+	if (effect->id >= T500RS_MAX_EFFECTS) {
+		hid_err(t500rs->hdev, "Effect ID %d exceeds maximum %d\n",
+			effect->id, T500RS_MAX_EFFECTS);
+		return -EINVAL;
+	}
+
+	/* Validate effect parameters based on type */
+	/* Per-type range checks. We only reject values that are genuinely
```
Kimplul, line 595: There are no range checks?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+	/* Device expects LITTLE-ENDIAN and value = range * 60. */
+	range_value = range * 60;
+
+	/* Send Report 0x40 0x11 [value_lo] [value_hi] to set range */
+	{
+		struct t500rs_pkt_r40_config *config =
+			(struct t500rs_pkt_r40_config *)buf;
+		config->id = 0x40;
+		config->subcmd = 0x11;
+		config->data1 = range_value &
```
Kimplul, line 930: `cpu_to_le16()` would be more clear here.

README.md
```diff
@@ -113,6 +113,15 @@ sudo yum install linuxconsoletools # Fedora-based
 > meaning that the wheels have to be initialized with `tmdrv`. Please see
 > https://github.com/Kimplul/hid-tmff2/issues/48.
 
+> **NOTE:** The T500 RS (like other modern Thrustmaster wheels) first enumerates
```
Kimplul, line 116: If the T500 is the same as other wheels, why add this note?

README.md
```diff
@@ -171,6 +180,11 @@ for wheels:
   ```shell
   echo 'blacklist hid_thrustmaster' | sudo tee /etc/modprobe.d/hid_thrustmaster.conf
   ```
+  **Only do this if this repo's `hid-tminit` is installed** (it is built and
```
Kimplul, line 183: This is not strictly related to the T500, I would suggest removing. Updating docs is something I've been thinking of, that can be handled separately.

src/tmt500rs/hid-tmt500rs.h
```diff
@@ -0,0 +1,157 @@
...
+ * Used internally by the sequencing system to manage packet order.
+ */
+enum t500rs_seq_packet {
+	T500RS_SEQ_ENVELOPE,
+	T500RS_SEQ_CONSTANT,
+	T500RS_SEQ_MAIN,
+};
+
+/* Supported effects */
+extern const signed short t500rs_effects[];
```
Kimplul, line 80: I don't think this needs to be exposed in the header.

src/tmt500rs/hid-tmt500rs.h
```diff
@@ -0,0 +1,157 @@
...
+ * - b2: effect type (T500RS_EFFECT_* constants)
+ * - b3: control flags (always 0x40)
+ * - b4-b5: duration in milliseconds (LE)
+ * - b6-b7: delay before start in milliseconds (LE)
+ * - b8: reserved (0x00)
+ * - b9-b10: parameter packet subtype (LE) - determines 0x03 codes
+ * - b11-b12: envelope packet subtype (LE) - determines 0x02 code
+ * - b13-b14: reserved (0x0000)
+ */
+struct t500rs_pkt_r01_main {
```
Kimplul, line 107: Similarly, these can probably just fine be in `hid-tmt500rs.c`. Generally, a header should only expose what's required to use the corresponding `*.c` file, which should be very little in this case. `hid-tmff2.c` doesn't need to know the packet structure of the wheels, just that a `t500rs_populate_api()` function exists :)

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+		// Keep explicit zeros even though memset() clears them.
+		config->data1 = 0x00;
+		config->data2 = 0x00;
+	}
+	ret = t500rs_send_hid(t500rs, init_buf, 4);
+	if (ret)
+		hid_warn(t500rs->hdev,
+			 "Autocenter-disable (0x40 0x04) failed: %d\n", ret);
+
+	/* Report 0x43 - Set global gain (2 bytes). Advisory: seed the device
```
Kimplul, line 1115: Advisory?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+	 */
+	memset(init_buf, 0, 2);
+	init_buf[0] = 0x43;
+	init_buf[1] = (u8)(clamp_t(int, default_gain, 0, 100) * 255 / 100);
+	ret = t500rs_send_hid(t500rs, init_buf, 2);
+	if (ret)
+		hid_warn(t500rs->hdev,
+			 "Initial gain set (0x43) failed: %d\n", ret);
+
+	hid_info(t500rs->hdev, "T500RS initialized successfully (HID mode)\n");
```
Kimplul, line 1131: Are there other modes?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+	 *
+	 * Note: 0x40 subcommand 0x11 is the RANGE command
+	 * (docs/T500RS_FFBEFFECTS.md section 5.7) and is never sent at init.
+	 */
+	{
+		struct t500rs_pkt_r40_config *config =
+			(struct t500rs_pkt_r40_config *)init_buf;
+		config->id = 0x40;
+		config->subcmd = 0x04;
+		// Keep explicit zeros even though memset() clears them.
```
Kimplul, line 1106: Why?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+		config->data1 = 0x00;
+		config->data2 = 0x00;
+	}
+	ret = t500rs_send_hid(t500rs, init_buf, 4);
+	if (ret)
+		hid_warn(t500rs->hdev,
```
Kimplul, line 1112: Actually, why doesn't this just call `t500rs_set_autocenter()` directly?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+
+	memset(init_buf, 0, 2);
+	init_buf[0] = 0x42;
+	init_buf[1] = 0x00;
+	ret = t500rs_send_hid(t500rs, init_buf, 2);
+	if (ret)
+		hid_warn(t500rs->hdev, "Init command 0x42 0x00 failed: %d\n",
+			 ret);
+
+	/* Defensive state reset: a module reload (rmmod/insmod) does not
```
Kimplul, line 1079: Man, this comment is verbose. Just say you're stopping any effects that may be lying around or something more to the point, this is rather tedious to try and parse.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+	/* Use send_buffer for all HID transfers */
+	init_buf = t500rs->send_buffer;
+
+	T500RS_DBG(t500rs, "Sending initialization sequence...\n");
+
+	/* Report 0x42 - Init/status commands (2 bytes each)
+	* Windows sends these at startup: 0x42 0x04, 0x42 0x05, 0x42 0x00
+	* These appear to initialize the FFB subsystem state.
+	*
+	* The opening sync (0x42 0x04) is mandatory: if the device cannot
```
Kimplul, line 1047: Is this not the `open` command that another comment claims the T500 doesn't use?

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+		return -EINVAL;
+	}
+
+	if (!tmff2->hdev || !tmff2->input_dev) {
+		pr_err("t500rs_wheel_init: Invalid tmff2 structure"
+		       " (missing hdev or input_dev)\n");
+		return -EINVAL;
+	}
+
+	hid_dbg(tmff2->hdev, "T500RS: Initializing HID mode\n");
```
Kimplul, line 993: Not using `T500RS_DBG` here for whatever reason. For what it's worth, I don't really see the benefit of the macro anyway, and would maybe prefer just using `hid_dbg` directly, but eh.

src/tmt500rs/hid-tmt500rs.c
```diff
@@ -0,0 +1,1216 @@
...
+	}
+
+	ret = t500rs_send_hid(t500rs, buf, 4);
+	if (ret) {
+		hid_err(t500rs->hdev, "Failed to send range command: %d\n",
+			ret);
+		goto out;
+	}
+
+	/* Apply settings with Report 0x42 0x05 */
```
Kimplul, line 943: Applying settings seems to be used a few times, would it make sense to have a separate function for it?

cazzoo:
> > Gain and range via the core parameters; autocentering routed through the core workqueue (T500RS callbacks do blocking HID I/O and must not run from atomic context).
> 
> This bit is the preparatory PR, and shouldn't really be included in this PR. You should be able to change the target of this PR to #220 instead of `master`, which splits off the commits that are not about the T500 constant force changes. That's the idea of stacked PRs anyway.

I am not sure about the stack. I tried to link them (it's linked locally on my PC using gh CLI) but when I attempt to push, I am having 422 errors:
`⚠ failed to update base branch for PR #221: HTTP 422: Validation Failed (https://api.github.com/repos/Kimplul/hid-tmff2/pulls/221)
PullRequest.base is invalid`

In the WebUI, If I edit the PR target, I cannot select the initial PR (with the defer), the reason being that my code resides in a fork, and this new PR cannot target the other PR itself. Either i'm doing wrong, either I just missed something.

src/tmt500rs/hid-tmt500rs.c
cazzoo: We did, indeed, but after the captures I got from community members, I figured that it's really true. ID 1 is reserved for constant, and the following ones for the effects

Kimplul: So what happens if you try to upload two constant force effects on Windows?

Kimplul:
Huh, weird. I've only used the GitLab version of stacked PRs, which worked as expected, and you could select the PR from the dropdown of target branches.

Anyway, no big deal if we can't figure this out right now, we can just focus on one PR at a time and when the lowest one gets merged, the commits should disappear from the others PR and the discussion hopefully stays rather more focused.

Kimplul:
Darn, read through the [GitHub docs about stacked PRs](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs) and noticed

> Stacked pull requests require all branches to be in the same repository. Cross-fork stacks are not supported.

That's annoying, I guess they're currently targeting more enterprise-y situations. Sorry, didn't realize earlier, but let's keep this informatl stack going, I suspect it's still better than one huge PR.

kakra:
Maybe just keep a local queue branch, then push only a reviewable small commit series. Wait for feedback. Create fixup commits, fold them into the PR commits, then rebase. If the PR is ready and merged, rebase your remaining work, then create a new PR.

